PR #267, opened 2012-07-25
This sets up the sqlite vs non-sqlite paths at setup time, rather than run time, which should have performance benefits.

Empact:
The tests run green but I wouldn't consider it well-veted just yet. I still need to run it against apps running & not running sqlite. Perhaps you can give that a go?


alexreisner:
Closing, because old and many conflicts, but still open to the idea.

PR #454, opened 2013-04-26
Handle Content-Encoding 'gzip' responses from Geocoding API's.

thrillcall:
Now requests gzip'ed responses by default for improved API performance.

```
Geocoder.config[:http_headers] = {"Accept-Encoding" => "gzip"}
```

This request behavior can be overridden with:

```
Geocoder.config[:http_headers] = nil
```

This fix is the result of an issue observed with Bing where it will
sometimes return the first response as:

```
"content-type"=>["application/json"]
```

with no content-encoding response header. However it will sometimes
return gzip body content on subsequent requests:

```
"content-type"=>["application/json"], "content-encoding"=>["gzip"]
```

with the body fully gzip encoded. Geocoder did not previously handle
gzip'ed responses.

### DIFF
--- a/lib/geocoder/configuration.rb
+++ b/lib/geocoder/configuration.rb
@@ -1,3 +1,5 @@
+# -*- encoding : utf-8 -*-
+
 require 'singleton'
 require 'geocoder/configuration_hash'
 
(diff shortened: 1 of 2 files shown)


alexreisner:
Closing due to lack of activity, but I'm still open to adding this. The :http_headers config param is already part of the gem, we just need to address the rescue issue (need specific exceptions).

PR #658, opened 2014-05-29
Allow to calculate the distance using a second set of coordinates

sildur:
I need to find all places near a specific location, but sorted by a second location, and showing the distance from such second location. I'm adding a parameter, select_location, which allows to use that second location to calculate distance, and hence, to sort by such distance.


alexreisner:
Closing because I think it adds too much complexity for what's really an edge case. This sorting can be accomplished in a couple of other ways (w/out addition of this feature).

PR #700, opened 2014-08-24
use hashed cache key to prevent too long cache key

dbourguignon:
We hit some time using the cache an error due to a too long cache key.

I changed to use the md5 of the url as a cache key.
Maybe this should be optional through configuration.

What do you think?


alexreisner:
Closing. For discussion please see #984.

PR #727, opened 2014-09-30
Add configuration to turn off warn/log messages

### LINKED ISSUE #721: The usage of warn in the rescue blocks
Can you remove the usage of warn and replace to something like puts? To be easily stubbed by rspec

I'm running rspec and it's a pain to silent the warn logger...
screenshot 2014-09-26 19 16 23

The present workaround its something like...

``` ruby
      before(:all) do
        module ::Kernel; alias_method :old_warn, :warn; def warn(msg); end; end
      end

      after(:all) do
        module ::Kernel; remove_method :warn; define_method :warn, :old_warn.to_proc;end
      end
```

Instead of simply

`$stdout.stub(:puts)`

letz:
Fix #721

Signed-off-by: Ricardo Leitão <letzdevelopment@gmail.com>

### DIFF
--- a/lib/geocoder/configuration.rb
+++ b/lib/geocoder/configuration.rb
@@ -55,7 +55,8 @@ module Geocoder
       :cache_prefix,
       :always_raise,
       :units,
-      :distances
+      :distances,
+      :log
     ]
 
     attr_accessor :data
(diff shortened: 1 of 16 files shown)


TrangPham:
@letz I know that it's been a long time since your pull request, but has #806 helped solved your issue? Wondering if that is enough for @alexreisner to close this PR. If there's anything more you require for logging, please let us know.

letz:
@TrangPham thanks for fixing this issue :+1:

PR #746, opened 2014-11-09
Geocoder fallback

rob-murray:
Hi alex,
I've had another go at implementing fallback options, as discussed in #661 and I have gone with your api - so you can configure fallbacks like this;

```ruby
:lookup => [
    {
        :name => :yandex,
        :skip => ->(query) { !query.text.match(/russia/i) },
        :failure => ->(results, exception) { exception.class == Geocoder::OverQueryLimitError }
    },
    {
        :name => :google,
        :skip => nil,
        :failure => nil
    }
  ]
```

If the fallback options arent specified, ie the `:lookup` is just the symbol then it works as before.

I am pretty much done, however I have only been testing this locally and not in the real world whereas the changes I had in #661 have been used in real applications for a while.

Anyway, welcome your thoughts so far.


rob-murray:
bump! 
Hi, I have updated this branch from your repo's master and removed we only have the fallback execution strategy now; if just the plain lookup symbol is specified then it builds a fallback type lookup that will never skip.
Let me know if you'd like anything else changed - it would be good to get something like this in the gem.

cassidyclawson:
Just another bump from a fan that would like to see this feature in master.

My personal use case would be failover from freegeoip.net to telize.

alexreisner:
@rob-murray @cassidyclawson (and everyone else who's interested): sorry for the delay. I've been on vacation the past two weeks and swamped with work upon returning, and given the large amount of code in this pull request it's going to take me some more time to review. Rest assured, though, I'm excited to see this and hope to start diving into it soon.

rob-murray:
No problem @alexreisner - If there is anything you'd like changed then let me know.

l4u:
:+1:

tmaszk:
Would love to see this changed merged

Mszniber:
Hi !
Still wondering if this got implemented or not. Any news ?

TrangPham:
@Mszniber - This is a feature that @alexreisner and I think is useful and important to Geocoder. I believe there requires more thoughts and brainstorm into a good solution.

Mszniber:
Need some help on that ?

PR #792, opened 2015-01-23
Handle freegoip's updated response format

hosamaly:
It seems that freegeoip.net now returns `zip_code` (with an underscore) instead of `zip_code`. This pull request handles both responses.

### DIFF
--- a/lib/geocoder/results/freegeoip.rb
+++ b/lib/geocoder/results/freegeoip.rb
@@ -29,7 +29,7 @@ module Geocoder::Result
     end
 
     def postal_code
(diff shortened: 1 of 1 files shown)


hosamaly:
I tried to write a test for it, but I didn't know how to do it while skipping the fixtures. Here is the integration test as I would have written it:

```ruby
def test_freegeoip_zipcode
  lookup = Geocoder::Lookup::Freegeoip.new
  result = lookup.search('2.30.95.202').first
  assert_match /E10.*/, result.address
end
```

Issue #753, opened 2014-11-19
Bing: raise error when HTTP header X-MS-BM-WS-INFO set to 1


TrangPham:
From http://msdn.microsoft.com/en-us/library/ff701703.aspx 
```
Responses that do not return results
Occasionally, the servers processing service requests can be overloaded, and you may receive 
some responses that contain no results for queries that you would normally receive a result. 
To identify this situation, check the HTTP headers of the response. If the HTTP header 
X-MS-BM-WS-INFO is set to 1, it is best to wait a few seconds and try again.
```

Would it be possible to add a check for the header X-MS-BM-WS-INFO and raise an error if this happens? I think this would be useful for checking if the request should be tried again later.

PR #794, opened 2015-01-27
Handle Bing over limit when header 'X-MS-BM-WS-INFO' is set to 1

TrangPham:
For #753 

@alexreisner Please let me know if anything needs to be changed for this. I wasn't 100% sure if the `Geocoder::OverQueryLimitError` was the correct error to be raising. Thanks~


alexreisner:
This looks great, thanks! Just a few things:

1. Can we create a new exception called `ServiceUnavailable` and raise that instead? We'd also raise that if, in theory, a lookup returned HTTP 503 (though I don't think any of them currently do).

2. Along similar lines, let's rename the response fixture `bing_service_unavailable`.

3. I think the value of the HTTP header in the real world will be a string (`"1"`). Can we confirm that and, if so, check for that instead?

4. I think you left a debugging line in the lookup (`puts response`).

TrangPham:
Thanks Alex! 

I think you're right about the HTTP header return `"1"` not `1` since it is a json. I'll double check that.
And I'll make the changes you suggested.

4. Oops :)

TrangPham:
@alexreisner For adding the case when 503 happens, do you mean add it for then the `response.status == 503` or when the `response.body["statusCode"] == "503"`?

Because I see that `results(query)` does some error handling using the `statusCode`. Based on what I've read (here: https://msdn.microsoft.com/en-us/library/ff701707.aspx), I think that `.status` and `statusCode` should have the same code, anyhow.

alexreisner:
I mean `response.code`. You don't actually need to do that as part of this PR, but if you want to, see `Lookup::Base#check_response_for_errors!`. I don't think Bing will ever return a 503, we'll just raise the same exception in this case where perhaps they should.

TrangPham:
I looked into how the header is returned. I see that `Lookup::Base.http_client` specifies `Net::HTTP`

net/http actually returns the response headers as shown below:
(tested in irb)
``` Ruby
> response.each_header{|h| puts h}
cache-control
transfer-encoding
content-type
server
x-bm-traceid
x-bm-srv
x-ms-bm-ws-info
x-aspnet-version
x-powered-by
date
connection
 => {"cache-control"=>["no-cache"], "transfer-encoding"=>["chunked"], "content-type"=>["application/json; charset=utf-8"], "server"=>["Microsoft-IIS/8.0"], "x-bm-traceid"=>["813072cc1abb442b82be509c344a580c"], "x-bm-srv"=>["HK20271643"], "x-ms-bm-ws-info"=>["0"], "x-aspnet-version"=>["4.0.30319"], "x-powered-by"=>["ASP.NET"], "date"=>["Tue, 27 Jan 2015 15:23:11 GMT"], "connection"=>["close"]}
> response["x-ms-bm-ws-info"]
 => "0"
```

So yes - it's a string not an int, but also the headers are always downcase and are not accessible from `response.headers["header-name"]` like I had thought but actually `response["header-name"]`

I see that in MockHttpResponse, there is a method
``` Ruby
  def [](key)
    send key if respond_to?(key)
  end
```
This is meant to return the headers, correct?

For this to work the initializer would have to dynamically make functions for the key such that the mock object would `respond_to?` it.
``` Ruby
   (options[:headers] || {}).each do |header_key, header_value|
      (class << self; self; end).class_eval do
        define_method header_key.downcase, Proc.new { header_value }
      end
    end
``` 
This would work. 

But I purpose that we instead change the `def [](key)` method like so:
``` Ruby
def [](key)
    @headers[key]
end
```
I'm not sure if the current definition of this function is being depended on by other code.

@alexreisner What do you think?

alexreisner:
Good questions. Net::HTTP's interface is so weird. I think you should change the implementation of `MockHttpResponse#[]` to what you suggest. If its goal is to mock a Net::HTTP response then I'd say the current implementation is incorrect.

TrangPham:
@alexreisner Ready to be reviewed again.

alexreisner:
Looks great! Thanks for the good discussion and the good code.

TrangPham:
Thanks for your great feedback :) 
This is my first merge into an open source project actually. I'd be happy to contribute more to this project. If there's a need, let me know.

PR #795, opened 2015-01-27
Added street, house and sub_city methods for Yandex result

wired9:


alexreisner:
Hmm...I think these methods are too specific to your application. In general I'm hesitant to add methods that are lookup-specific like this.

Issue #776, opened 2014-12-23
Handling IP address-like ZIP code


kenn:
We run a ZIP code search using Google:

```ruby
Geocoder.search(zip_code, params: { components: "country:#{country}"})
```

Today, we got an exception because a ZIP code "4.1.0.2" was entered by a user (the country was Philippines) and recognized as an IP address, and processed with `Freegeoip`, which doesn't have the `types` method in a result as Google does.

I don't think "4.1.0.2" is a normal format people would type in (I assume that there's no IPv4-like ZIP code format exists in the world.), and I don't want to waste my time investigating it again in the future.

Possible solutions:

- Set `:ip_lookup => :google` config - The user probably get a random result from Google (with above case, [this](http://maps.googleapis.com/maps/api/geocode/json?address=4.1.0.2&components=country%3APH&language=en&sensor=false)), but at least I won't get bogus notification on `NoMethodError`.
  - or create a pull request to add an option to disable IP lookup - the consequence is same as `:ip_lookup => :google`
- Monkey patch the `ip_address?` method and raise a specific error that can be used to turn into a validation error - more desirable result.
  - or create a pull request to add an option to raise a specific error.

Anyone experienced this, and if so, what's your recommendation?

alexreisner:
This should be solved by #798 and 0a0d359.

Issue #351, opened 2012-12-08
Multiple Lookup Strategy


alexreisner:
We need a way to effectively use multiple geocoding APIs. Situations that need to be handled:
- The primary API is unavailable and we'd like to automatically fall back to another.
- The primary API is returning bad/unexpected results (or quota is reached) and we'd like to automatically fall back to another.
- We'd like to use different APIs for different queries, depending on whether the query is an IP address or a physical address, but also if the query is for a certain part of the world where one API is significantly better than another (eg: Yandex in Russia).

Ideas for implementation:
1. Set :lookup to an array of preferred APIs (descending preference).
2. Set :lookup to a block which takes a query and returns a lookup name, or an array of names (as in #1).

Need to determine:
- How to "fall back" to a less-preferable API. On timeout? On any API-related exception? On any Geocoder exception? On a particular API response? On a `nil` result? And how configurable does it need to be?
- How to handle IP vs. physical address lookups in the config. Currently they are configured separately (:lookup and :ip_lookup). Is there an advantage to combining them?

darmenise:
Hello @alexreisner I'm loving this gem and, referencing your third bullet point, I need to use different APIs for different queries. In particular, I'd like to use LocationIQ for the front-end because it's cheap but inaccurate, and Google for the admin back-end because it's expensive but very accurate. In other words: 

_One **cheap service** for lots of queries that doesn't need too much precision and one **expensive service** for a few queries that require great precision._

At the current state of the gem, how can one achieve this please?

It would be ideal being able to specify which strategy/api to use for each lookup, defaulting to the [first configured service](https://github.com/alexreisner/geocoder#configuring-multiple-services).

FeSens:
Here is how I implemented an easy fallback strategy:
```
  geocoded_by :full_address, lookup: lambda{ |obj| obj.geocoder_lookup }

  def geocoder_lookup
    if Geocoder.search(full_address, lookup: :nominatim).present?
      :nominatim
    else
      :falllback_geocoder (:google, :esri, etc..)
    end
  end
```

Why it works:
The geocoder_lookup will try for your first and then second option. If it get anything the results are cached by the Geocoder on Redis. This makes it cheap for a second lookup when the `geocode` is called again.

PR #798, opened 2015-01-30
Force specfify ip_lookup on search, even when ip_address? fails

TrangPham:
Connected to #776

### DIFF
--- a/lib/geocoder/query.rb
+++ b/lib/geocoder/query.rb
@@ -32,7 +32,7 @@ module Geocoder
     # appropriate to the Query text.
     #
     def lookup
-      if ip_address? || (options[:ip_lookup] && options[:lookup].nil?)
+      if options[:ip_address] || ip_address?
         name = options[:ip_lookup] || Configuration.ip_lookup || Geocoder::Lookup.ip_services.first
       else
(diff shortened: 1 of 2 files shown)


alexreisner:
Thanks for tackling this problem. I like the simplicity of this solution, but I think it should be even simpler. The issue is that we eventually want to be able to configure multiple lookups (see #351 and #746) and that configuration should be kept to the `Geocoder.configure` method. Even regardless of that, we shouldn't need to re-specify our IP lookup when all we really want is to be able to force a particular bit of input to be treated as an IP address. Right? What if we just check for the truthiness of `options[:ip_address]`? I also think that should be checked before the `ip_address?` method is run, as checking the hash value is a faster operation than running the method.

TrangPham:
@alexreisner Thanks I've made the changes that you suggested.

alexreisner:
Thanks! Also, I realized after I merged that you can use more specific assertions in your tests. See e065f6de.

TrangPham:
@alexreisner Thanks

The same change from assert_equal  to assert_instance_of can also be made to `query_test.rb` line 50, and in `geocoder_test.rb` line 64, 70, 76 for consistency

alexreisner:
Good eye!

mmzoo:
Sweet! Thank you very much.

Issue #594, opened 2014-02-05
geocoder needs to support geoip2 from Maxmind


bevanhunt:
geoip is no longer available for purchase from maxmind and it is the version that geocoder supports. 

http://dev.maxmind.com/geoip/geoip2/whats-new-in-geoip2/

alexreisner:
This was added by #800.

PR #799, opened 2015-01-30
Update readme for :maxmind to specify that it is the legacy web services...

TrangPham:
... not the GeoIP2 Precision Service

Connected to #594

Just to make it more clear which service Geocoder actually supports (since the old link actually now points to the GeoIP2 Precision page)


alexreisner:
Excellent, thanks.

PR #800, opened 2015-02-02
Adding MaxMind's GeoIP2 Precision Service API

TrangPham:
Connected to #594 

I know that geocoder currently has a geoip2 lookup defined which uses Hive::Geoip2 

- though I believe this is the lite version using the free database? 
- should I be adding this functionality to :geoip2 directly and have some logic that understands to use the Precision Service API?

I ran the test cases, but locally `bundle exec rake test` seems to indicate that the test_helper's version of MaxmindGeoip2 might not actually be used 

- Do I have something configured wrong for the tests?

### DIFF
--- a/lib/geocoder/configuration.rb
+++ b/lib/geocoder/configuration.rb
@@ -55,7 +55,8 @@ module Geocoder
       :cache_prefix,
       :always_raise,
       :units,
-      :distances
+      :distances,
+      :basic_auth
     ]
 
     attr_accessor :data
@@ -96,6 +97,7 @@ module Geocoder
       @data[:api_key]      = nil         # API key for geocoding service
       @data[:cache]        = nil         # cache object (must respond to #[], #[]=, and #keys)
       @data[:cache_prefix] = "geocoder:" # prefix (string) to use for all cache keys
+      @data[:basic_auth]   = {}          # user and password for basic auth ({:user => "user", :password => "password"})
 
       # exceptions that should not be rescued by default
       # (if you want to implement custom error handling);
--- a/lib/geocoder/lookup.rb
+++ b/lib/geocoder/lookup.rb
@@ -56,7 +56,8 @@ module Geocoder
         :maxmind,
         :maxmind_local,
         :telize,
-        :pointpin
+        :pointpin,
+        :maxmind_geoip2
       ]
     end
 
--- a/lib/geocoder/lookups/base.rb
+++ b/lib/geocoder/lookups/base.rb
@@ -271,8 +271,8 @@ module Geocoder
           opts[:use_ssl] = use_ssl?
 
           http_client.start(*args, opts) do |client|
-            req = client::Get.new(uri.request_uri, configuration.http_headers)
-            if configuration.basic_auth
+            req = http_client::Get.new(uri.request_uri, configuration.http_headers)
+            if configuration.basic_auth[:user] && configuration.basic_auth[:password]
               req.basic_auth(configuration.basic_auth[:user], configuration.basic_auth[:password])
             end
             client.request(req)
--- a/lib/geocoder/lookups/base.rb
+++ b/lib/geocoder/lookups/base.rb
@@ -271,7 +271,11 @@ module Geocoder
           opts[:use_ssl] = use_ssl?
 
           http_client.start(*args, opts) do |client|
-            client.get(uri.request_uri, configuration.http_headers)
+            req = http_client::Get.new(uri.request_uri, configuration.http_headers)
+            if configuration.basic_auth
+              req.basic_auth(configuration.basic_auth[:user], configuration.basic_auth[:password])
+            end
+            http_client.request(req)
           end
         end
       end
--- a/lib/geocoder/lookups/base.rb
+++ b/lib/geocoder/lookups/base.rb
@@ -271,11 +271,11 @@ module Geocoder
           opts[:use_ssl] = use_ssl?
 
           http_client.start(*args, opts) do |client|
-            req = http_client::Get.new(uri.request_uri, configuration.http_headers)
+            req = client::Get.new(uri.request_uri, configuration.http_headers)
             if configuration.basic_auth
               req.basic_auth(configuration.basic_auth[:user], configuration.basic_auth[:password])
             end
-            http_client.request(req)
+            client.request(req)
           end
         end
       end
--- a/lib/geocoder/lookups/maxmind_geoip2.rb
+++ b/lib/geocoder/lookups/maxmind_geoip2.rb
@@ -15,7 +15,7 @@ module Geocoder::Lookup
     end
 
     def query_url(query)
-      "#{protocol}://geoip.maxmind.com/geoip/v2.1/#{configured_service!}/" + url_query_string(query)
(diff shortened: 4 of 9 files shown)


alexreisner:
Thanks for this. Let's address these two items and then revisit the test failures:

1. You don't need to modify Lookup::Base#make_api_request. If you look at the top of that method you'll see the username and password being extracted from the parsed URI. Just generate a query URL with the username and password included (eg: http://user:pass@geoip.maxmind.com/...).
2. You don't need to add `:basic_auth` to the base configuration hash. Each lookup will have its own credentials so the developer should do that configuration for this lookup only, as you do correctly in `test_helper.rb` (no Geocoder code modification needed).

TrangPham:
@alexreisner 
About not modifying `Lookup::Base#make_api_request` - I wasn't too sure about the whole basic_auth verses proxy. I thought they were different. Or is it that the proxy can be used instead of using basic authentication?

alexreisner:
Hmmm...I believe that syntax is just how basic auth works. I don't think it implies a proxy.

TrangPham:
For the format that you mentioned: ` http://user:pass@geoip.maxmind.com/`
`Lookup::Base#http_client` parses it using URI and creates a Net::HTTP::Proxy which is then used to run the get

The GeoIP2 Precision Web Services points to http://en.wikipedia.org/wiki/Basic_access_authentication for more info about basic auth.

alexreisner:
Look at [line 104](https://github.com/TrangPham/geocoder/blob/geoip2/lib/geocoder/lookups/base.rb#L104). It only creates a proxy if `:http_proxy` or `:https_proxy` is configured.

alexreisner:
[Line 269](https://github.com/TrangPham/geocoder/blob/geoip2/lib/geocoder/lookups/base.rb#L269) is where user/pass are extracted from the URI. On lines 274 and 278, in the `http_client.start` block, you should be using `client`, not `http_client` (which may be the source of the test failures), but again, I don't think this code should be modified.

TrangPham:
@alexreisner Ah okay I was slightly confused.

I did a simple test with Net::HTTP using http://httpbin.org/
``` Ruby
$ irb
2.2.0 :001 > require 'net/http'
 => true
2.2.0 :003 > uri = URI.parse("http://httpbin.org/basic-auth/user1/passwd2")
 => #<URI::HTTP http://httpbin.org/basic-auth/user1/passwd2>
2.2.0 :017 > args = [uri.host, uri.port, nil, nil, "user1", "passwd2"]
 => ["httpbin.org", 80, nil, nil, "user1", "passwd2"]
2.2.0 :018 > client = Net::HTTP.start(*args, {})
 => #<Net::HTTP httpbin.org:80 open=true>
2.2.0 :021 > req = Net::HTTP::Get.new(uri.request_uri)
 => #<Net::HTTP::Get GET>
2.2.0 :022 > client.request(req)
 => #<Net::HTTPUnauthorized 401 UNAUTHORIZED readbody=true>
2.2.0 :023 > req.basic_auth("user1", "passwd2")
 => ["Basic dXNlcjE6cGFzc3dkMg=="]
2.2.0 :024 > client.request(req)
 => #<Net::HTTPOK 200 OK readbody=true>
```

It seems that basic authentication with net/http can only be done through the request.basic_auth method, and not the client args.

alexreisner:
Okay, thanks, that does seem like a good proof. Just be sure to use `client` instead of `http_client` within the `start` block, or get rid of the `start` block if it's not necessary. Also remove the `args`-related lines if they're not necessary. But make sure to preserve SSL support!

TrangPham:
@alexreisner Yes thanks I have made sure to use the `client` instead of `http_client` within that `start` block. 

I found out the issue I was having with failing tests was because the `query_url` that I had for maxmind_geoip2 was wrong (i.e. did not generate a unique url so that the caching could work properly). Thus the same result was being grabbed for each call. 

I looked back at the documentation for GeoIP2 and the url should be:
`https://geoip.maxmind.com/geoip/v2.1/{service}/{ip_address}`
where in service is one of city, country, or insights

I have made those changes to the url in maxmind_geoip2.

I have also manually tested geocoder a little to ensure the the basic_auth addition to make_api_request did not break everything.

``` Ruby
Thus-MacBook-Air:geocoder trangpham$ gem build geocoder.gemspec
  Successfully built RubyGem
  Name: geocoder
  Version: 1.2.7
  File: geocoder-1.2.7.gem
Thus-MacBook-Air:geocoder trangpham$ gem install ./geocoder-1.2.7.gem
Successfully installed geocoder-1.2.7
Parsing documentation for geocoder-1.2.7
Done installing documentation for geocoder after 1 seconds
1 gem installed
Thus-MacBook-Air:geocoder trangpham$ irb
2.1.2 :001 > require 'geocoder'
 => true
2.1.2 :002 > Geocoder.configure(timeout: 10, lookup: :bing, units: :km, api_key: "MY_API_KEY_IS_HIDDEN")
 => {:timeout=>10, :lookup=>:bing, :ip_lookup=>:freegeoip, :language=>:en, :http_headers=>{}, :use_https=>false, :http_proxy=>nil, :https_proxy=>nil, :api_key=>"MY_API_KEY_IS_HIDDEN", :cache=>nil, :cache_prefix=>"geocoder:", :basic_auth=>{}, :always_raise=>[], :units=>:km, :distances=>:linear}
2.1.2 :003 > Geocoder.search("San Francisco, CA")
 => [#<Geocoder::Result::Bing:0x007faeb567b568 @data={"__type"=>"Location:http://schemas.microsoft.com/search/local/ws/rest/v1", "bbox"=>[37.64891052246094, -122.66365814208984, 37.92567443847656, -122.18305206298828], "name"=>"San Francisco, CA", "point"=>{"type"=>"Point", "coordinates"=>[37.77711868286133, -122.41963958740234]}, "address"=>{"adminDistrict"=>"CA", "adminDistrict2"=>"San Francisco Co.", "countryRegion"=>"United States", "formattedAddress"=>"San Francisco, CA", "locality"=>"San Francisco"}, "confidence"=>"High", "entityType"=>"PopulatedPlace", "geocodePoints"=>[{"type"=>"Point", "coordinates"=>[37.77711868286133, -122.41963958740234], "calculationMethod"=>"Rooftop", "usageTypes"=>["Display"]}], "matchCodes"=>["Good"]}, @cache_hit=nil>, #<Geocoder::Result::Bing:0x007faeb567b0b8 @data={"__type"=>"Location:http://schemas.microsoft.com/search/local/ws/rest/v1", "bbox"=>[37.78520584106445, -122.40876007080078, 37.80015563964844, -122.39292907714844], "name"=>"Financial District, CA", "point"=>{"type"=>"Point", "coordinates"=>[37.79383087158203, -122.40196990966797]}, "address"=>{"adminDistrict"=>"CA", "adminDistrict2"=>"San Francisco Co.", "countryRegion"=>"United States", "formattedAddress"=>"Financial District, CA", "locality"=>"Financial District"}, "confidence"=>"High", "entityType"=>"Neighborhood", "geocodePoints"=>[{"type"=>"Point", "coordinates"=>[37.79383087158203, -122.40196990966797], "calculationMethod"=>"Rooftop", "usageTypes"=>["Display"]}], "matchCodes"=>["Good"]}, @cache_hit=nil>]
```

TrangPham:
Thanks @alexreisner

alexreisner:
Thank you! I merged this and made a few minor changes. This is the most substantive change: 7331bd7d.

Also, see #801.

PR #804, opened 2015-02-08
Improve request.location(), add request.paranoid_location()

edslocomb:
The original impetus for this was to fix exceptions thrown due to HTTP headers set by cruddy proxies, e.g.:

* an exception would be thrown when HTTP_X_FORWARDED_FOR was set to ","

Other problems noticed along the way, and fixed:

* `location` is vulnerable to trivial IP spoofing via HTTP headers; this has been added to the docs, and an alternate but less accurate `paranoid_location` provided.

* `location` did not filter out proxies trusted by Rack, so for people running Rack behind their own proxy (as recommended, but frequently ignored), `location` would always return the location of that proxy.  `location` and `paranoid_location` now respect Rack/Rails trusted_proxies config

* there's a whole zoo of nonstandard HTTP headers used by proxies to store the upstream address, a few more of them are checked now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,12 +123,14 @@ The exact code will vary depending on the method you use for your geocodable str
 Request Geocoding by IP Address
 -------------------------------
 
-Geocoder adds a `location` method to the standard `Rack::Request` object so you can easily look up the location of any HTTP request by IP address. For example, in a Rails controller or a Sinatra app:
+Geocoder adds `location` and `paranoid_location` methods to the standard `Rack::Request` object so you can easily look up the location of any HTTP request by IP address. For example, in a Rails controller or a Sinatra app:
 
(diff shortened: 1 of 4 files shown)


alexreisner:
This is a great pull request. Thanks for the thorough testing and documentation. The only question I have is whether the new method should be called `safe_location` instead of `paranoid_location` since one need not be paranoid to use it, only prudent. I think it makes the same point, and perhaps more clearly. Any thoughts on that?

edslocomb:
Oh, I have no attachment to the method name at all.  By all means, change it to whatever fits your house style; I just went with a convention I'd used on other projects.

alexreisner:
OK, cool. I renamed the method. Thanks!

PR #806, opened 2015-02-11
Adding Geocoder.log() 

TrangPham:
In response to #727 

I have followed the specifications from http://www.ruby-doc.org/stdlib-2.1.0/libdoc/logger/rdoc/Logger.html for the case when the user passes in a Logger to config

### DIFF
--- a/lib/geocoder.rb
+++ b/lib/geocoder.rb
@@ -1,4 +1,5 @@
 require "geocoder/configuration"
+require "geocoder/logger"
 require "geocoder/query"
 require "geocoder/calculations"
 require "geocoder/exceptions"
--- a/lib/geocoder/configuration.rb
+++ b/lib/geocoder/configuration.rb
@@ -99,7 +99,7 @@ module Geocoder
       @data[:cache]        = nil         # cache object (must respond to #[], #[]=, and #keys)
       @data[:cache_prefix] = "geocoder:" # prefix (string) to use for all cache keys
       @data[:basic_auth]   = {}          # user and password for basic auth ({:user => "user", :password => "password"})
-      @data[:logger]       = :default    # nil, default, or Logger instance
+      @data[:logger]       = :kernel     # :kernel or Logger instance
 
       # exceptions that should not be rescued by default
       # (if you want to implement custom error handling);
--- a/lib/geocoder/configuration.rb
+++ b/lib/geocoder/configuration.rb
@@ -56,7 +56,8 @@ module Geocoder
       :always_raise,
       :units,
       :distances,
-      :basic_auth
+      :basic_auth,
+      :logger
     ]
 
     attr_accessor :data
@@ -98,6 +99,7 @@ module Geocoder
       @data[:cache]        = nil         # cache object (must respond to #[], #[]=, and #keys)
       @data[:cache_prefix] = "geocoder:" # prefix (string) to use for all cache keys
       @data[:basic_auth]   = {}          # user and password for basic auth ({:user => "user", :password => "password"})
+      @data[:logger]       = :default    # nil, default, or Logger instance
 
       # exceptions that should not be rescued by default
       # (if you want to implement custom error handling);
--- a/lib/geocoder/logger.rb
+++ b/lib/geocoder/logger.rb
@@ -18,17 +18,19 @@ module Geocoder
     }
 
     def log(level, message)
+      return nil unless valid_level?(level)
+
       logger = Geocoder.config[:logger]
-      return true unless valid_level?(level)
 
       if logger == :kernel
         kernel_log(level, message)
-      elsif logger.respond_to? :add
+      elsif logger.kind_of? ::Logger
         logger.add(SEVERITY[level], message)
       else
         raise Geocoder::ConfigurationError, "Please specify valid logger for Geocoder. " +
         "Logger specified must be :kernel or must respond to `add(level, message)`."
       end
+      nil
     end
 
     private
--- a/lib/geocoder/logger.rb
+++ b/lib/geocoder/logger.rb
@@ -1,0 +1,50 @@
+require 'logger'
+
+module Geocoder
+
+  def self.log(level, message)
+    Logger.instance.log(level, message)
+  end
+
+  class Logger
+    include Singleton
+
+    SEVERITY = {
+      debug: Logger::Severity::DEBUG,
+      info: Logger::Severity::INFO,
+      warn: Logger::Severity::WARN,
+      error: Logger::Severity::ERROR,
+      fatal: Logger::Severity::FATAL
+    }
+
+    def log(level, message)
+      logger = config[:logger]
+      return true unless logger && valid_level?(level)
+
+      if logger == :default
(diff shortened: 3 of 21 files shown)


alexreisner:
This is great! Thanks. Comments:

1. If you set `logger: Rails.logger` (as certainly will be many people's first instinct) you get errors because the default ActiveSupport logger (and many others) don't have a `#log` method. Instead they use `#warn`, `#error`, etc. So I think we should add some `respond_to?`s before selecting a method to call on the logger object.
2. If you set `logger: nil` (which is the default) it doesn't work (does not log). Should either fix this or change the code comment in Configuration (I don't think we need to support `nil`).
3. You missed an instance of `warn` in the `Base` lookup.

Also, I wonder if it should be `logger: :kernel` instead of `logger: :default`. It's more descriptive.

TrangPham:
@alexreisner 

According to the code source ActiveSupport::Logger is a subclass of ::Logger
https://github.com/rails/rails/blob/6329d9fa8b2f86a178151be264cccdb805bfaaac/activesupport/lib/active_support/logger.rb#L44

And has `.log` as shown.
``` Ruby
2.2.0 :004 > l = ActiveSupport::Logger.new('log.txt')
 => #<ActiveSupport::Logger:0x007fa05cb7ffa8 @progname=nil, @level=0, @default_formatter=#<Logger::Formatter:0x007fa05cb7fee0 @datetime_format=nil>, @formatter=#<ActiveSupport::Logger::SimpleFormatter:0x007fa05cb7fb48 @datetime_format=nil>, @logdev=#<Logger::LogDevice:0x007fa05cb7fe40 @shift_size=1048576, @shift_age=0, @filename="log.txt", @dev=#<File:log.txt>, @mutex=#<Logger::LogDevice::LogDeviceMutex:0x007fa05cb7fdc8 @mon_owner=nil, @mon_count=0, @mon_mutex=#<Mutex:0x007fa05cb7fd28>>>>
2.2.0 :005 > l.level = ::Logger::WARN
 => 2
2.2.0 :006 > l.add(::Logger::WARN, "test")
 => true
2.2.0 :007 > File.read('log.txt')
 => "# Logfile created on 2015-02-16 14:05:52 -0800 by logger.rb/47272\ntest\n"
2.2.0 :008 > l.log(::Logger::WARN, "test")
 => true
2.2.0 :009 > File.read('log.txt')
 => "# Logfile created on 2015-02-16 14:05:52 -0800 by logger.rb/47272\ntest\ntest\n"
2.2.0 :010 > l.log(::Logger::INFO, "test2")
 => true
2.2.0 :011 > File.read('log.txt')
 => "# Logfile created on 2015-02-16 14:05:52 -0800 by logger.rb/47272\ntest\ntest\n"
```

But I'll change it to `.add` since ::Logger#log is an alias to it and ActiveSupport::Logger#add overrides it

I'll remove `nil` as an option.

Also for the :kernel versus :default. I agree that it's more descriptive, but I thought it might be confusing? Looking at it again I agree that it should be changed to :kernel

alexreisner:
Wait, sorry. I think we can stay with `#log`. You're right about `ActiveSupport::Logger`. Let me re-examine the cause of this error...

alexreisner:
I found the issue. It happens on older versions of Rails. For example, on 3.0.9 (which we still need to support) I get this:

    NoMethodError: undefined method `log' for #<ActiveSupport::BufferedLogger:0x007fc65b5de278>
        from /Users/alex/dev/geocoder/lib/geocoder/logger.rb:27:in `log'
        from /Users/alex/dev/geocoder/lib/geocoder/logger.rb:6:in `log'
        from /Users/alex/dev/geocoder/lib/geocoder/lookups/base.rb:186:in `rescue in parse_json'
        from /Users/alex/dev/geocoder/lib/geocoder/lookups/base.rb:180:in `parse_json'
        from /Users/alex/dev/geocoder/lib/geocoder/lookups/base.rb:193:in `parse_raw_data'
        from /Users/alex/dev/geocoder/lib/geocoder/lookups/base.rb:169:in `fetch_data'
        from /Users/alex/dev/geocoder/lib/geocoder/lookups/mapquest.rb:40:in `results'
        from /Users/alex/dev/geocoder/lib/geocoder/lookups/base.rb:47:in `search'
        from /Users/alex/dev/geocoder/lib/geocoder/query.rb:11:in `execute'
        from /Users/alex/dev/geocoder/lib/geocoder.rb:21:in `search'
        from (irb):1

So, at some point Rails' logging mechanism changed...

TrangPham:
Can you see if `add` works on the older versions of Rails?

alexreisner:
Aha! Yes, `#add` does appear to work. So you were right about that too...

TrangPham:
Ah also `ActiveSupport::BufferedLogger` does have a depreciated warning:
http://edgeapi.rubyonrails.org/classes/ActiveSupport/BufferedLogger.html

alexreisner:
Yes, I believe the change was for [twelve-factor compliance](http://12factor.net/logs) or at least resulted from a similar philosophy.

TrangPham:
I've changed some of the `warn`s that I've missed

alexreisner:
It's looking good. One issue I'm seeing, though, is that `Geocoder.log` returns different values depending on which underlying logger is used. `Kernel#warn` always returns `nil`, while `ActiveSupport::Logger#add` returns the message logged as a string. This causes different behavior if, for example, invalid JSON is received. Most lookups depend on the return value of `fetch_data` to determine whether results were returned. If invalid JSON is returned, a warning is logged in the base lookup but the value is `nil`/`false` if `logger: :kernel` and `"Geocoder's response..."`/`true` if `logger: Rails.logger`. This leads to an uncaught exception in several cases.

To duplicate, try setting `lookup: :mapquest` and setting an invalid key.

I think the solution here is as simple as making `parse_json` return false in the rescue block but I think we should go back and look for other places where a truthy (rather than falsy) return values might cause different results.

On a somewhat related note, why does `Geocoder.log` return `true` if an invalid level is given? I'm not saying this is wrong, I'm just curious.

TrangPham:
@alexreisner 

I decided to return true since the the `::Logger` returns true when the level is not high enough: http://ruby-doc.org//stdlib-2.1.0/libdoc/logger/rdoc/Logger.html#method-i-add

Though now that you mention it, it seems more like a raised exception might be better behaviour?

I'll have to look into the different return values.

TrangPham:
@alexreisner 

Ah okay - I think a good solution here is to make the `Geocoder.log` always return nil - this is the expected behaviour of what happened before.

The reason why this error is happening is because fetch_data will now return `true` instead of nil when something is logged. The lookup's `results` functions already handle the case that the fetch_data is nil. 

I've also changed Logger.log to check if `logger.kind_of? ::Logger` instead of checking `logger.respond_to? :add`. 

```Ruby
$ rails console
 > ::Logger.new("Log.txt").kind_of? ::Logger
 => true
 > Logger.new("Log.txt").kind_of? ::Logger
 => true
```

alexreisner:
Looks great. Thanks for the sample console output!

PR #810, opened 2015-02-24
add support for deCarta geocoder

tyrasd:
This adds support for the [deCarta](http://developer.decarta.com/Docs/REST/Search) geocoding service.


TrangPham:
Hi @tyrasd 

I have a couple comments on your work. 

1. I don't think that you need to overwrite `fetch_data` and `parse_raw_data` in your lookup. It can just rely on base lookup.

2. Could you add a test file for your lookup? 
For a reference you could look at `yahoo_test.rb`. If you need any help writing your tests or understanding `test_helper.rb`, I would be glad to answer your questions.

TrangPham:
@tyrasd Thank you for your PR!

It has been inactive for 11 months and is not in a mergeable state. Thus I will be closing it.

Please let us know if you wish to continue working on it. We welcome more contributions from you.

Issue #83, opened 2011-07-10
support for datamapper


kwando:
Hey! This gem is wonderful, it would be nice if it supported datamapper as a ormlayer though.
//Hannes

heliohead:
+1

alexreisner:
Sorry, DataMapper support is not going to be added. Please see #811 for explanation. Also see the `Geocoder::Sql` module for help in working with DataMapper.

PR #811, opened 2015-02-25
Adding support for DataMapper as a model

TrangPham:
Connected to #83 

Please note that I have currently refrained from implementing heavier SQL dependant functions such as `.near`, due to lack of knowledge on DataMapper.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ source "https://rubygems.org"
 group :development, :test do
   gem 'rake'
   gem 'mongoid', '2.6.0'
+  gem 'data_mapper', '1.2.0'
   gem 'bson_ext', platforms: :ruby
   gem 'geoip'
   gem 'rubyzip'
--- a/lib/geocoder.rb
+++ b/lib/geocoder.rb
@@ -10,6 +10,7 @@ require "geocoder/ip_address"
 require "geocoder/models/active_record" if defined?(::ActiveRecord)
 require "geocoder/models/mongoid" if defined?(::Mongoid)
 require "geocoder/models/mongo_mapper" if defined?(::MongoMapper)
+require "geocoder/models/data_mapper" if defined?(::DataMapper)
 
 module Geocoder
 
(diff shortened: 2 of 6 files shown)


alexreisner:
Thanks for this. Unfortunately I don't think we're going to add support for DataMapper. I should have closed that old issue with a comment to that effect. The focus of Geocoder is on the process of geocoding. The `near` method (and other query methods) are provided as a convenience to support the most common use cases for lat/lon data. The `Geocoder::Sql` module is provided for anyone who wants to implement distance queries using another object-relational mapper (see the `full_distance` method for implementing `near`).

Sorry about this. I'm going to close #83.

alexreisner:
@TrangPham Thanks so much for all of your recent work on Geocoder. It is _very_ much appreciated! If you're interested in thinking about database-related problems, you might be interested in tackling one of the biggest and longest-standing problems in Geocoder: testing database queries. Currently ActiveRecord is stubbed and they are simply not tested. I am honestly not sure of the best way to test them and have not thought about this problem in a couple of years, but it really needs to be solved.

If you're interested, I think the first step is opening an issue where we can discuss and fully describe the problem. Ideally we want to run real queries on a real database, and do so in multiple versions of rails (3.0.x, 3.1.x, 3.2.x, 4.0.x, 4.1.x, 4.2.x) and against multiple database types (PostgreSQL, MySQL, SQLite, MongoDB). I've used the [geocoder_test](https://github.com/alexreisner/geocoder_test) project for running these tests in the past but it's out of date and I don't think the approach is very good anyway. I'm not sure how other gems handle this kind of testing. Maybe what I'm proposing is overkill, but we've had enough Rails version- and DB type-specific bugs over the years that I think this level of detail is warranted.

Having a more complete test suite will also make me feel more comfortable adding DB-related features, as requested by many of the old PRs (eg, DataMapper support).

This is a big task and might be more than you want to take on. Obviously, feel free to contribute whatever you like.

TrangPham:
@alexreisner I'm interested, though I would need your input on the direction you would want it to go. Can you create the issue for it and we'll talk more on there about what specific steps need to be taken to actualize testing like that?

Thank you

PR #812, opened 2015-02-26
add more results to google and nominatim

twetzel:
add **district** and **subdistrict** with **_code** to *google* and *nominatim* results

**more google additions:**
- street
- city_code

**more nominatim additions:**
- street
- city
- street_number
- country_code
- city_code

### DIFF
--- a/lib/geocoder/results/google.rb
+++ b/lib/geocoder/results/google.rb
@@ -99,6 +99,61 @@ module Geocoder::Result
       @data['address_components']
     end
 
+
+    ## Additional Stuff
+    def street
+      if route = address_components_of_type(:route).first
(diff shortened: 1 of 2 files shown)


alexreisner:
Thanks for this, but I think if people want this level of detail they should dig into the results themselves. The problem is that these methods amount to basically renaming of address components, but standardization of these components is impossible. "District" means very different things in different places (it's basically meaningless in the US). These methods mask the complexity of the underlying data and will lead to confusion.

twetzel:
yeah ok .. I just use them with german addresses and even there subdistrict is just available in a handfull of cities
(but I saw there is neighborhood in the results so I thought it may help)

PR #818, opened 2015-03-05
Enable testing different database environiments

TrangPham:
Summary:
Enabled running tests using `rake test DB=`, with options mysql, sqlite, and postgres. If DB is not set (or is nil) the tests will run as before (using mocked ActiveRecord::Base).

Also configured to run them on travis-ci

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ rdoc/*
 .bundle
 Gemfile.lock
 api_keys.yml
+test/geocoder_test.sqlite
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,9 @@
 language: ruby
+env:
+  - DB=
+  - DB=sqlite
+  - DB=postgres
+  - DB=mysql
 rvm:
   - 1.9.3
   - 2.0.0
--- a/Gemfile
+++ b/Gemfile
@@ -7,6 +7,7 @@ group :development, :test do
   gem 'geoip'
   gem 'rubyzip'
   gem 'rails'
+  gem 'sqlite3'
   gem 'test-unit' # needed for Ruby >=2.2.0
 
   gem 'byebug', platforms: :mri
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,6 @@ group :development, :test do
   gem 'geoip'
   gem 'rubyzip'
   gem 'rails'
-  gem 'sqlite3'
   gem 'test-unit' # needed for Ruby >=2.2.0
 
   gem 'byebug', platforms: :mri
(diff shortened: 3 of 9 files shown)


TrangPham:
@alexreisner What I've decided to do is update the Rakefile such that `rake test DB=db_type` is now possible.

Right now I have postgres running pretty well. (As shown in top comment).
I will be working on having mysql and sqlite3 also working locally.

Please let me know what you think about this approach? Thanks!

TrangPham:
@alexreisner 

I've got sqlite, postgres, and mysql running (failing some of the SQL specific tests though).
You can see that they run on travis-ci now! 

There's still a bit of clean up and refactoring to do.

alexreisner:
**This is awesome!!** I'm really excited to see DB tests running on Travis. It seems like the hard part is already done. If you can get all DB tests to pass I'll merge this and then we can start adding tests from the `geocoder_test` repo separately. Thank you for the great work!

TrangPham:
@alexreisner I've fixed the failures :smiley: But for sqlite - there are some tests in `NearTest` that say `without_sqlite`, are these not meant to pass when using sqlite?

alexreisner:
Yes, that's correct. Skipping those tests on SQLite is the right thing to do. The only thing I'm wondering is if it's better to use `omit`, eg:

    omit("Not applicable to SQLite") if ENV['DB'] == 'sqlite'

as the first line for each of those tests. Will it still be considered passing on Travis?

TrangPham:
@alexreisner It should be (that's how it behaves locally).

alexreisner:
Wow, so annoying. Even though [1.9.3 is no longer officially supported](https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/) it's too soon to just drop it from Geocoder. I think the failure can be fixed by adding `minitest` to the 1.9.3 `Gemfile` and calling `skip` instead of `omit` if Ruby version is <2.0. Probably a good idea to add a test helper method like `omit_if_sqlite` and handle the logic in there.

Once that's done I think this is ready for merge.

And thanks again. This is huge.

TrangPham:
I think it was just that `test-unit` gem was missing from the Gemfile for 1.9.3. It seems to be working once I added that.

TrangPham:
I added the `test-unit` gem to the Gemfile for 1.9.3 and that seems to make the tests pass - though I'm not sure if that's the right way to go about it. Since it will be installing an updated version (now 3.0.9), whereas the included Test::Unit in ruby was 1.2.3 (according to the read me: https://github.com/test-unit/test-unit)

alexreisner:
Ah, I think that's fine. Merging now. Amazing to have this done! Thanks for the great work!

PR #819, opened 2015-03-11
use builtin libs to validate ips

astropanic:
Ruby has already good Regex for checking IP addresses 
```Resolv::IPv4::Regex``` and ```Resolv::IPv6::Regex```

No need to reinvent the wheel.

666.65.123.94 - technically or not, this address is not valid


alexreisner:
Awesome! Thanks!

astropanic:
you're welcome ;)

Issue #816, opened 2015-03-04
`Geocoder::ServiceUnavailable` is not implemented in 1.2.7


malandrina:
`Geocoder::ServiceUnavailable` is referenced in the README but doesn't appear to have been implemented in the 1.2.7 release. It looks like it was implemented a few days after the release: https://github.com/alexreisner/geocoder/commit/49647fa32c4b63c543fc3e7727bc10d1a288a9aa

To avoid confusion, I'd propose removing the reference in the README or updating the README to clarify that `Geocoder::ServiceUnavailable` is only available on `master`. I'd be happy to open a PR to make this change.

alexreisner:
This raises a question I've thought about periodically but have never really dealt with. Because of the way Github works, people view the README on master/HEAD as if it's the definitive README for the latest version of the gem. But it isn't. It's for master/HEAD. In the case you mention, the README is technically correct, you're just looking at the wrong version of it (probably everyone is), because that's the version Github shows by default.

I'm not sure what to do about this. Putting README changes in a separate branch would solve it, but it would also put code and documentation in separate commits/merges, which I don't like (should be able to revert code/doc changes together).

Does that make sense? Any thoughts on that are very welcome.

Also, related to what you point out, note that only a few of the exceptions available in Geocoder are raised by any given lookup, so there's no guarantee that if you configure Geocoder to raise `ServiceUnavailable` that it will actually be raised under those conditions (because most APIs don't return 503 when they should...you may get a timeout instead). Perhaps this is something that should be explained in the README.

malandrina:
> Does that make sense? Any thoughts on that are very welcome.

Yes, definitely! I agree that it’s best to be able to revert code and documentation changes together. I have some thoughts on different approaches I’ve seen.

1. `ActiveModel::Serializers` maintains branches for different releases and has an [enormous, front-and-center disclaimer](https://github.com/rails-api/active_model_serializers#maintenance-please-read) about this in the README. This acts as a useful reminder that the docs you’re viewing on GitHub may not be relevant to the version you’re using on your project.
2. The [`paperclip` README](https://github.com/rails-api/active_model_serializers#maintenance-please-read) explicitly states the following:

> Also, everything in this README might not apply to your version of Paperclip, and you should read the README for version 2.7 instead.

3. The [`factory_girl` README](https://github.com/thoughtbot/factory_girl#documentation) doesn’t include documentation, and instead instructs users to reference documentation “for your version of factory_girl” on Rubygems.

[...]

PR #820, opened 2015-03-13
Update background job example

malandrina:
* Per discussion from https://github.com/alexreisner/geocoder/issues/816,
  this commit replaces the `sidekiq_worker` example with an example
  using ActiveJob.
* Demonstrate approach to reverse-geocoding records in a background job
  with simple retry logic for timeout and socket errors.


alexreisner:
Thanks! This is good, but I think it can be more general if `perform` takes an object instead of an ID (possible thanks to [GlobalID serialization](http://edgeguides.rubyonrails.org/active_job_basics.html#globalid)). This would prevent the need for hard-coding the `Location` class and would even enable us to take it out of `/examples` and make it available to apps via something like:

    Geocoder::Jobs::ReverseGeocode.perform_later(location)

malandrina:
Makes sense -- thanks! Updated in https://github.com/alexreisner/geocoder/commit/f00546ef764e469789e4000e7ec13c050719e4f4.

alexreisner:
Looks great! Thanks!

PR #823, opened 2015-03-21
Don't include test files in the exported gem

TrangPham:
Will make the exported gem size smaller.

Before: 1.1M
screen shot 2015-03-21 at 10 32 28 am

After: 516k
screen shot 2015-03-21 at 10 43 12 am

Note: size found using `bundle list geocoder` then `du -h` on the file path it gave.


alexreisner:
Nice, thanks. I never liked having `git ls-files` in there, but it used to be standard (Bundler boilerplate, I believe) back in the day.

If all we're including in the gem is executable/library code (plus README/LICENSE/CHANGELOG) then maybe we don't need `/examples` either?

TrangPham:
I was looking at rails/activerecord gemspec and some other gems. They seem to leave things like the examples and guides type of things in

alexreisner:
Alright, sounds good to me!

alexreisner:
Version 1.2.8 just released. Smallest payload since 1.1.8. And your name is [all over it](https://github.com/alexreisner/geocoder/blob/master/CHANGELOG.md). Thanks for all the great work!

Issue #836, opened 2015-04-07
geoip2 and maxminddb do not work


zekus:
the maxminddb result set do not support the fetch method. The latest version of the maxminddb gem includes a "to_hash" method that perhaps is sueful to make the resultset compatible with the current geoip2 implementation again.
```ruby
[1] pry(main)> g = Geocoder.search "205.189.0.33"
[2] pry(main)> g.first.country
NoMethodError: undefined method `fetch' for #<MaxMindDB::Result:0x00000103ede6b8>
from /Users/zekus/.rvm/gems/ruby-2.1.2@archisnapper/gems/geocoder-1.2.8/lib/geocoder/results/geoip2.rb:36:in `country'
```

TrangPham:
Thanks zekus. 

The change in maxminddb gem which caused this was: https://github.com/yhirose/maxminddb/commit/7b902ece6023d1844474802cbc663855ac33de81

Wherein `Hash(raw)` changed to `raw || {}`

`.fetch` is a method on Hash. Assuming that it worked previously to v0.1.7 of maxminddb gem. Then all that needs to be added is `to_hash` on `@data`. `to_hash` on a hash returns a hash, so this will support older and current versions of maxminddb gem as well as working with Hive::Geoip2.

alexreisner:
Closing, thanks to #837.

PR #837, opened 2015-04-08
Support maxminddb gem v0.1.7

TrangPham:
See #836


zekus:
@TrangPham I would rather add a method

```ruby
def data
  @data.to_hash
end
```

and use that in all the other methods where it refers to @data

TrangPham:
Thanks @zekus I've refactored as you've suggested.

Issue #802, opened 2015-02-04
Silence Test Warnings


alexreisner:
Currently several tests are producing irrelevant warnings. These should be silenced for clean test output.

PR #839, opened 2015-04-10
Silence Test Warnings

TrangPham:
for #802

### DIFF
--- a/lib/geocoder/lookups/mapquest.rb
+++ b/lib/geocoder/lookups/mapquest.rb
@@ -15,8 +15,7 @@ module Geocoder::Lookup
 
     def query_url(query)
       domain = configuration[:open] ? "open" : "www"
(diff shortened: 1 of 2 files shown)


alexreisner:
This ensures silence for all tests, but we probably want to allow unexpected warnings, which can provide helpful information. I think what we really want to do here is silence only those warnings that are expected and caused intentionally in a test, almost like an `assert_raise` block. Does that make sense?

Using `silence_warnings` blocks (eg, see [test/unit/error_handling_test.rb](https://github.com/alexreisner/geocoder/blob/master/test/unit/lookups/yahoo_test.rb#L16)) in tests that produce warnings is probably sufficient for now.

TrangPham:
Ah okay. Thanks @alexreisner. 

I'll revert and use `silence_warnings`.

Issue #813, opened 2015-02-27
add ORM testing (in multiple environments) to test suite


TrangPham:
See the comment on #811 

We want to be able to test against multiple versions of Rails and different databases. "Ideally we want to run real queries on a real database, and do so in multiple versions of rails (3.0.x, 3.1.x, 3.2.x, 4.0.x, 4.1.x, 4.2.x) and against multiple database types (PostgreSQL, MySQL, SQLite, MongoDB)"

TrangPham:
@alexreisner I've created an issue so that we can start the discussion. Please feel free to change the title and top comment as you see fit.

I was looking into possibly using Travis CI to run different test environments. Travis CI has support for this as mentioned here: http://docs.travis-ci.com/user/languages/ruby/#Testing-against-multiple-versions-of-dependencies-(Ruby-on-Rails%2C-EventMachine%2C-etc) What do you think of this solution?

alexreisner:
Yes, Travis is already being used to test against all supported Ruby versions and that is definitely the approach we want to take for Rails versions as well. This may all be as simple as un-stubbing ActiveRecord (loading the real thing), creating some test objects (ideally using FactoryGirl), and running database queries (using `near` and other methods).

TrangPham:
Agreed - I think it would be simpler than originally thought. 

Would the testing be done in the geocoder project itself or would we be updating the geocoder_test project?

alexreisner:
The tests in the geocoder_test project are the ones we should start with. Would love to move them over to Geocoder and get rid of geocoder_test.

TrangPham:
@alexreisner I was able to run the tests in the geocoder_test/rails30, though some of them are failing for me. I think the next step would be to get geocoder_test to run in travis-ci and then to have it run with multiple gemfile configurations.

What do you think about that?

alexreisner:
Yeah, some of the tests fail. And it's partly for that reason that I think the best first step is actually to get just one test that queries the DB running in the main repository. (Then we can selectively bring in the tests from geocoder_test, and we may end up discarding some of them.) How does that sound?

TrangPham:
Okay that sounds like a plan :)

TrangPham:
@alexreisner What are the next steps for this issue? To migrate over the tests that are in `geocoder_test/base/test`?

alexreisner:
[...]

PR #845, opened 2015-04-21
Testing against diff versions of rails

TrangPham:
Work in Progress
-------------------------
Connected to #813 

There are very few test failures so far with rails 4.1 gemfile.

The mongoid one is due to:
```
#3037 Model indexes are no longer stored in an index_options hash on the model class. Instead, an array named index_specifications now exists on the class which contains a list of Indexable::Specification objects. This is so we could properly handle the case of indexes with the same keys but different order.
```
Non-backward compatible change in Mongoid 4.0. I think this is only in our test case that `index_options` is used.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,10 +14,20 @@ rvm:
 gemfile:
   - Gemfile
   - gemfiles/Gemfile.ruby1.9.3
+  - gemfiles/Gemfile.rails3.2
+  - gemfiles/Gemfile.rails4.1
 matrix:
   exclude:
     - rvm: 1.9.3
       gemfile: Gemfile
+    - rvm: 1.9.3
+      gemfile: gemfiles/Gemfile.rails3.2
+    - rvm: 1.9.3
+      gemfile: gemfiles/Gemfile.rails4.1
+    - env: DB=
+      gemfile: gemfiles/Gemfile.rails3.2
+    - env: DB=
+      gemfile: gemfiles/Gemfile.rails4.1
     - rvm: 2.0.0
       gemfile: gemfiles/Gemfile.ruby1.9.3
     - rvm: 2.1.2
--- a/gemfiles/Gemfile.rails3.2
+++ b/gemfiles/Gemfile.rails3.2
@@ -1,0 +1,38 @@
+source "https://rubygems.org"
+
+group :development, :test do
+  gem 'rake'
+  gem 'mongoid', '2.6.0'
+  gem 'bson_ext', platforms: :ruby
+  gem 'geoip'
+  gem 'rubyzip'
+  gem 'rails', '>= 3.2'
+  gem 'test-unit' # needed for Ruby >=2.2.0
+
+  gem 'byebug', platforms: :mri
+
+  platforms :jruby do
+    gem 'jruby-openssl'
+    gem 'jgeoip'
+  end
+
+  platforms :rbx do
+    gem 'rubysl', '~> 2.0'
+    gem 'rubysl-test-unit'
+  end
+end
+
+group :test do
+  gem 'sqlite3', :platform => [:ruby, :mswin, :mingw]
+
+  platforms :ruby do
+    gem 'pg'
+    gem 'mysql2'
+  end
+
+  platforms :jruby do
+    gem 'jdbc-mysql'
+    gem 'jdbc-sqlite3'
+    gem 'activerecord-jdbcpostgresql-adapter'
+  end
+end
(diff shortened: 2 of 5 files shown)


alexreisner:
The test failures are unrelated to this commit. I fixed them with 899a34b521458. However, it looks like you've specified the Rails version incorrectly:

    gem 'rails', '>= 3.2'

Won't that just load the latest version of Rails?

TrangPham:
I think you're right - should be `~>`

Thanks for fixing the unrelated failures!

alexreisner:
[The documentation](http://guides.rubygems.org/patterns/#declaring-dependencies) is annoyingly unclear about this, but I believe you actually want:

    gem 'rails', '~> 3.2.0'

...or 3.2.1 or 3.2.8, but I think the squiggly arrow means the least-valued digit in the version number can be anything larger than what's specified, so that `~> 3.1` would allow 3.2.x, which is undesirable.

However, I may be wrong about this. I've never seen a clear and complete definition of `~>`.

TrangPham:
Hmmm this is what I did to test the version of rails that would get installed:

``` Ruby
trangpham@geocoder:~/workspace (master) $ export BUNDLE_GEMFILE=$PWD/gemfiles/Gemfile.rails3.2
trangpham@geocoder:~/workspace (rails) $ bundle install
Fetching gem metadata from https://rubygems.org/.........
Fetching additional metadata from https://rubygems.org/..
Resolving dependencies...
Using rake 10.4.2
Using i18n 0.7.0
Using multi_json 1.11.0
Installing activesupport 3.2.21
Installing builder 3.0.4
Installing activemodel 3.2.21
Using erubis 2.7.0
Installing journey 1.0.4
Installing rack 1.4.5
Installing rack-cache 1.2
Using rack-test 0.6.3
Using hike 1.2.3
Using tilt 1.4.1
Installing sprockets 2.2.3
Installing actionpack 3.2.21
Installing mime-types 1.25.1
Installing polyglot 0.3.5
Installing treetop 1.4.15
Installing mail 2.5.4
Installing actionmailer 3.2.21
Installing arel 3.0.3
Installing tzinfo 0.3.44
Installing activerecord 3.2.21
Installing activeresource 3.2.21
Installing bson 1.12.1
Installing bson_ext 1.12.1
Using bundler 1.7.6
Installing columnize 0.9.0
Installing byebug 4.0.5
Installing geoip 1.5.0
Using json 1.8.2
Installing mongo 1.12.1
Installing mongoid 2.6.0
Installing mysql2 0.3.18
Installing pg 0.18.1
Installing power_assert 0.2.3
Installing rack-ssl 1.3.4
Installing rdoc 3.12.2
Using thor 0.19.1
Installing railties 3.2.21
Installing rails 3.2.21
Installing rubyzip 1.1.7
Using sqlite3 1.3.10
Installing test-unit 3.0.9
Your bundle is complete!
Use `bundle show [gemname]` to see where a bundled gem is installed.
Post-install message from rdoc:
Depending on your version of ruby, you may need to install ruby rdoc/ri data:

<= 1.8.6 : unsupported
 = 1.8.7 : gem install rdoc-data; rdoc-data --install
 = 1.9.1 : gem install rdoc-data; rdoc-data --install
>= 1.9.2 : nothing to do! Yay!
trangpham@geocoder:~/workspace (rails) $ 
```

`Installing rails 3.2.21` Seems to be what it does.

alexreisner:
That looks like what we want. What did you use in the `Gemfile`?

TrangPham:
I left the gemfile as `>=3.2`

Looking carefully at the documentation you;ll see it says 
```
# gemspec
spec.add_runtime_dependency 'library',
  '>= 2.2.0'

# bundler
gem 'library', '>= 2.2.0'

This is an “optimistic” version constraint. It’s saying that all changes from 2.x on will work with my software, but for version 3.0.0 this will not be true.

The alternative here is to be “pessimistic”. This explicitly excludes the version that might break your code.

# gemspec
spec.add_runtime_dependency 'library',
  ['>= 2.2.0', '< 3.0']

# bundler
gem 'library', '>= 2.2.0', '< 3.0'
```

Meaning that it would only vary the numbers after the major build number.

alexreisner:
OK, that makes sense, but according to that definition won't `>= 4.1` install 4.2.x?

TrangPham:
You're right - I checked and the gemfile4.1 would install 4.2.1

TrangPham:
@alexreisner Hi - I went on hiatus for quite a while. I want to check in to see what is left to do for this PR.
The tests were running on my travis-ci: https://travis-ci.org/TrangPham/geocoder

alexreisner:
Welcome back! This looks done. For some reason I neglected to merge.

PR #863, opened 2015-05-25
Stop :baidu and :baidu_ip trying to use SSL

matthewrudy:
The URL is fixed to http, so if `configuration.use_https?` is set to `true` it breaks.

This implements the same `hack` as `telize` uses.

Namely

``` ruby
# http only
def use_ssl?
  false
end
```

### DIFF
--- a/lib/geocoder/lookups/baidu.rb
+++ b/lib/geocoder/lookups/baidu.rb
@@ -18,6 +18,11 @@ module Geocoder::Lookup
 
     private # ---------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


matthewrudy:
I've added some test cases to the integration test,
but I don't know if you have baidu keys in your local `api_keys.yml` file.

If there's a better way we can specify this,
it'd be good.

Perhaps a lookup class can do:

```
self.supported_protocols = [:http]
```

and do it that way.

PR #864, opened 2015-05-25
`lookup.supported_protocols`

matthewrudy:
A single function to override on a `Lookup` class
that makes it clear which lookups support which protocols.

This is an alternative fix for
https://github.com/alexreisner/geocoder/pull/863


alexreisner:
Thanks for this! Closing #863.

PR #866, opened 2015-05-27
Updated Bing offering

guymorita:
Bing stats are outdated. They changed the offering sadly.


alexreisner:
Thanks! This is good, but can we use the "Compare plans" URL for "Key signup" instead?

guymorita:
they're a bit different. compare plans leads to key signup though.

alexreisner:
Yes, but every other entry in the README has "Key signup" and the address currently given there for Bing is out of date.

guymorita:
@alexreisner gotcha. updated the readme and replaced the old link. new link allows the user to compare plans first before signing up.

PR #873, opened 2015-06-13
logger configuration is incompatible with Rails.logger (not really)

costa:
The type checking seems redundant.


alexreisner:
Quack quack thanks.

PR #876, opened 2015-06-23
Fix yandex geocoder redirect to https

donbobka:
I added supported_protocols method with https only.


alexreisner:
For reference, their announcement about the switch to HTTPS:
http://clubs.ya.ru/mapsapi/replies.xml?item_no=56858

ZuevEvgenii:
@alexreisner Thank you!

Issue #678, opened 2014-07-16
Not seeing any Geocoder activity in my development.log.  As expected?


rreynier:
I am running tail -f on my development.log but I am not seeing any Geocoder activity. I am currently running rake jobs to geocode.  Is this expected behavior?  It would be really nice to see geocoding requests going out and coming back in.

alexreisner:
Logging was just added by #806.

PR #879, opened 2015-06-26
I believe this might be a leftover from development that someone forg…

LukeLaupheimer:
…ot to delete. Reject if I'm incorrect.

### DIFF
--- a/lib/geocoder/lookups/base.rb
+++ b/lib/geocoder/lookups/base.rb
@@ -273,7 +273,6 @@ module Geocoder
       def make_api_request(query)
         timeout(configuration.timeout) do
           uri = URI.parse(query_url(query))
-          Geocoder.log(:debug, "Geocoder: HTTP request being made for #{uri.to_s}")
           http_client.start(uri.host, uri.port, use_ssl: use_ssl?) do |client|
             req = Net::HTTP::Get.new(uri.request_uri, configuration.http_headers)
             if configuration.basic_auth[:user] and configuration.basic_auth[:password]


alexreisner:
Nope, it's intentional: c557ba436eadf750aa609bcef0010ca91d0f8f25. Requested feature.

LukeLaupheimer:
That's fine, but this thing dumps text right in the middle of my rspec tests. Example:

```
++ bundle exec rspec spec /Users/enablelabs/.rvm/gems/ruby-2.2.1@mori_pages/gems/respec-0.8.3/lib/respec/formatter.rb

Randomized with seed 63352
...........................................................................................................................................Geocoder: HTTP request being made for http://maps.googleapis.com/maps/api/geocode/json?address=415+River+Street%2C+Troy%2C+NY%2C+12180&language=en&sensor=false
...............................................................................................Geocoder: HTTP request being made for http://maps.googleapis.com/maps/api/geocode/json?address=415+River+Street%2C+Troy%2C+NY%2C+12180&language=en&sensor=false
..............................................................................

Finished in 4 minutes 21 seconds (files took 3.06 seconds to load)
312 examples, 0 failures
```

It's not like the gem is broken, but I will say this is annoying.

alexreisner:
Should only be appearing if log level is `:debug`, no?

LukeLaupheimer:
Nope, it goes direct to puts:
```ruby
# /lib/geocoder/logger.rb

# Line 38
def kernel_log(level, message)
  case level
  when :debug, :info
    puts message
```

alexreisner:
I think the correct solution to this is for the "kernel" logger to have a configurable level (default: `:warn`). I'd rather add that than remove the request URL logging, which is useful. (cc: @TrangPham)

LukeLaupheimer:
I mean, I suppose it could be arguably useful in some context maybe, but it's definitely not useful in mine. In fact, it's actively causing a problem. Normal use of a gem like this absolutely should not be outputting directly to `stdout`.

alexreisner:
Yes, it's useful. See #678. It's just shouldn't be on by default. Will fix soon (see #881).

Issue #886, opened 2015-07-16
Geocoder turning the coordenates


arturictus:
Using mongoid 
model:

```
self.coordinates = [latitude, longitude]
```

test:

``` ruby
let(:train_station){ create(:train_station, latitude: 53.34608, longitude: -6.296544) }
```

url generated:

```
GET http://maps.googleapis.com/maps/api/geocode/json?language=en&latlng=-6.296544,53.34608&sensor=false
```

byebug

```
[13, 22] in /Users/arturpanach/projects/gems/geocoder/lib/geocoder/lookups/google.rb
   13:       "http://maps.google.com/maps?q=#{coordinates.join(',')}"
   14:     end
   15:
   16:     def query_url(query)
   17:       byebug
=> 18:       "#{protocol}://maps.googleapis.com/maps/api/geocode/json?" + url_query_string(query)
   19:     end
(byebug) query
#<Geocoder::Query:0x007fe218d29388 @text=["-6.29698", "53.34608"], @options={:lookup=>nil, :language=>nil}>
```

FOUND
lib/geocoder/stores/mongo_base.rb:48

``` ruby
  ##
    # Coordinates [lat,lon] of the object.
    # This method always returns coordinates in lat,lon order,
    # even though internally they are stored in the opposite order.
    #
    def to_coordinates
      coords = send(self.class.geocoder_options[:coordinates])
      coords.is_a?(Array) ? coords.reverse : []
    end
```

**reverse?!!**

alexreisner:
See #832. Part of the reason for dropping Mongo support is that [Mongo stores the coordinates backwards](http://docs.mongodb.org/manual/applications/geospatial-indexes/#spherical) (hence `coords.reverse`) which causes all kinds of confusion, like this.

arturictus:
Hi I added support for mongoid-geospacial gem.
Tell me if That's enough to keep mongo support.
#887

nofxx:
Hello guys, before all ty for geocoder @alexreisner , use it on all projects.
Yeah, I'm doing the reverse with mongo. Thing is, mongo is actually correct:
It stores as lon,lat [x,y], but we talk (and google uses) latitude, longitude [y,x].

@arturictus anything I can help with, just pop a issue on the project. ty.

PR #887, opened 2015-07-16
fix reversing coordinates

arturictus:


arturictus:
Linked to #886

arturictus:
Added support for mongoid-geospacial gem.

alexreisner:
Thanks for this, and sorry for the very delayed response. This seems reasonable but we need a test or two showing that it works as expected.

alexreisner:
Closing, because while I like the idea of this, I think in an effort to simplify it may actually do the opposite. Generally I'm leaning towards reducing Geocoder's interaction with Mongo queries and data (since Mongo has its own way of doing geospatial operations).

PR #889, opened 2015-07-27
Safer timeouts

ankane:
Hi, thanks for the awesome work on this gem!

Ruby's timeout API is [inherently unsafe](http://www.mikeperham.com/2015/05/08/timeout-rubys-most-dangerous-api/). This PR replaces it with a safe method.

Open timeouts raise `Net::OpenTimeout` and read timeouts raise `Net::ReadTimeout`.  Both inherit from `Timeout::Error` (which is also defined as `TimeoutError`), so this should be a safe change, but we could always catch and re-raise a `Timeout::Error` you think that's better.

Let me know your thoughts. Thanks!

### DIFF
--- a/lib/geocoder/exceptions.rb
+++ b/lib/geocoder/exceptions.rb
@@ -29,4 +29,7 @@ module Geocoder
   class ServiceUnavailable < Error
   end
 
+  class LookupTimeout < ::Timeout::Error
+  end
+
 end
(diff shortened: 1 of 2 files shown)


alexreisner:
Excellent! Thanks for this! As for the exception, Geocoder should really raise a `Geocoder::LookupTimeout` so that future changes to the underlying timeout mechanism are hidden, but this exception class doesn't exist yet. What do you think is the safest way to introduce it? Is there a way to use the class hierarchy so that existing code that rescues from `Timeout::Error` continues to work?

ankane:
We can use the commit above, but the exception class won't be able to inherit from `Geocoder::Error` like the others.

alexreisner:
Awesome, thanks!! I think that is the best solution.

Issue #881, opened 2015-07-01
Kernel Logger Should Have Configurable Level


alexreisner:
Currently, the `:kernel` logger does not have a configurable level. Hence any `:debug` or `:info` messages are written to STDOUT. This is causing issues like #879. The kernel logger should have a configurable level and the default should be `:warn`.

alexreisner:
Fixed by the above commit (0bb6501).

PR #901, opened 2015-08-24
Never default to Puts for all debug messages

CharlesP:
This creates a mess of all terminal output. This should not be the default behavior.


CharlesP:
Using puts by default creates a mess of all terminal output. This should not be the default behavior.

alexreisner:
Sorry about this. Please see #881.

PR #908, opened 2015-09-09
Fix #906 - Use different memoization variables for request

### LINKED ISSUE #906: Security risk when calling both location & safe_location due to memoization?
Hello,

if you call request.location before request.safe_location, since they both use @location for memoization, request.safe_location will return the same value request.location did.

sowenjub:
See #906 safe_location and location should not rely on the same variable,
otherwise calling one before the other will yield unexpected results
and could even cause a security risk if location is called before
safe_location


alexreisner:
Great, thanks!

PR #909, opened 2015-09-09
Add locale support to GeoIP2 result display class

davydovanton:
:cat:

### DIFF
--- a/lib/geocoder/configuration.rb
+++ b/lib/geocoder/configuration.rb
@@ -125,6 +125,5 @@ module Geocoder
       end
       EOS
     end.join("\n\n"))
-
   end
 end
--- a/lib/geocoder/results/geoip2.rb
+++ b/lib/geocoder/results/geoip2.rb
@@ -13,35 +13,35 @@ module Geocoder
       end
 
       def latitude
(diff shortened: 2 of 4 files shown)


alexreisner:
Thanks! This is good, but could you add a test?

davydovanton:
@alexreisner hello, sorry, I hadn't time for this PR. I add test case for checking locale :smile:

alexreisner:
OK, thanks! A few things:

1. Take a look at the test failures on Travis. I think you just need to add `# encoding: utf-8` to the top of `test/test_helper.rb`.
2. Instead of `Geocoder.config[:language]`, please use `configuration[:language]` instead, which will return language for the specific lookup you're in.
3. In the test you added, if it fails the final line (which sets the language back to English) will not be executed. I think you need an `ensure`.

davydovanton:
@alexreisner thanks for feedback!
1. Yes, I loosed this, thanks for catch!
3. Instead `ensure` I use the `teardown` callback :smiley:

davydovanton:
all tests passed :blush:

alexreisner:
Looks good. Thanks!

davydovanton:
Thanks for merge and review! :tada:

PR #911, opened 2015-09-18
#656 rake task for mongo

mveer99:
mongo does not support find_each, replaced it with each


alexreisner:
This looks fine for Mongo but I think we still need to use `find_each` for ActiveRecord. `take` will only load the first `batch` records.

PR #913, opened 2015-09-21
Minor typo fix in the README

mehlah:
:scissors: "reverse" as the paragraph is only about simple geocoding. The next one is about reverse geocoding.


TrangPham:
:+1:

PR #917, opened 2015-09-29
Pass in dir variable when retreiving data files in Maxmind insert task

rayzeller:
Can't use a custom-located CSV when importing the maxmind DB.  This is an issue for Heroku users.


alexreisner:
Whoops! Thanks!

Issue #739, opened 2014-10-30
Support borough's


allcentury:
I recently answered a question on Stacko [here](http://stackoverflow.com/questions/26643768/finding-boroughs-with-geocoder/26644145#26644145) where the OP was asking if there was support for borough's.   

The use case seems small but I thought I'd ask if there's ever been a point where you've considered this?  I would assume for 99% of calls this will return nothing but for some it might be valuable.  I'm happy to give this a go if it's a feature the team is interested in. 

edit: I just noticed Alex is in Brooklyn!  (runs to double check code)

PR #919, opened 2015-10-01
Add county accessor for map quest results

Devetzis:
Trivial change to expose the County in the mapquests results. 

Test cases run and I've tested the gem in my own application.

### DIFF
--- a/lib/geocoder/results/mapquest.rb
+++ b/lib/geocoder/results/mapquest.rb
@@ -26,6 +26,10 @@ module Geocoder::Result
       @data['adminArea3']
     end
 
+    def county
+      @data['adminArea4']
+    end
+
     alias_method :state_code, :state
 
     #FIXME: these might not be right, unclear with MQ documentation


alexreisner:
The problem with this is that counties don't exist in all (most?) countries. (For discussion see #739, #362.)

Devetzis:
Alex:

I’ve read the referenced documents and I  don’t disagree, however: county is currently available for nominatim and opencagedata and sub_state is available from google and all are exposed by geocoder.

Additionally the map quest api http://open.mapquestapi.com/geocoding/ <http://open.mapquestapi.com/geocoding/> describes adminArea4 as County name and I’ve tested different locations and I get valid results (at least where I’ve checked).

So I understand your concern, but isn’t this also an issue for the geocode services I mentioned above?

Thanks…….

Perry Devetzis
Partner
Athena Technologies Corp.
perryd@athenatech.ca <mailto:perryd@athenatech.ca>
(403) 861-7556

Devetzis:
Alex:

Oh, it it’s only the method name ‘county’ I have no problems with ‘sub_state’.


Perry Devetzis
Partner
Athena Technologies Corp.
perryd@athenatech.ca <mailto:perryd@athenatech.ca>
(403) 861-7556

alexreisner:
OK, thanks for looking into this. Since it's mentioned in the API's docs I feel much better about it. Merging.

Issue #903, opened 2015-08-28
Stubbing invalid address coordinates request


sromocki:
Hello!

I'm looking to stub an invalid address for the Geocoder.coordinates(address) call.
If I call:

```point = Geocoder.coordinates("sadjdasks")```

point returns as nil

In order to stub a request like this I have tried the following code:

```
Geocoder::Lookup::Test.add_stub("sadjdasks", [])  # returns with error, undefined 'each_key'
Geocoder::Lookup::Test.add_stub("sadjdasks", [{}]) # returns [0.0,0.0]
Geocoder::Lookup::Test.add_stub("sadjdasks", nil) # returns with error, undefined 'map'
```

Perhaps, within results/test.rb you can check to see if data exists.  If not, then return nil.

Thanks!

alexreisner:
Great point. Open to a pull request. I think the correct way to specify no results should be:

    Geocoder::Lookup::Test.add_stub("sadjdasks", []) # empty array

as that is what `Lookup::Base#results` returns.

TrangPham:
@sromocki Hi I tried to reproduce your issue using ruby 2.1.2 and geocoder 1.2.11

This is the result I got:
```ruby
$ irb
2.1.2 :001 > require 'geocoder'
 => true
2.1.2 :003 > Geocoder.configure(:lookup => :test)
 => {:timeout=>3, :lookup=>:test, :ip_lookup=>:freegeoip, :language=>:en, :http_headers=>{}, :use_https=>false, :http_proxy=>nil, :https_proxy=>nil, :api_key=>nil, :cache=>nil, :cache_prefix=>"geocoder:", :basic_auth=>{}, :logger=>:kernel, :kernel_logger_level=>2, :always_raise=>[], :units=>:mi, :distances=>:linear}
2.1.2 :004 > Geocoder::Lookup::Test.add_stub("1", [])
 => []
2.1.2 :005 > Geocoder::Lookup::Test.add_stub("2", [{}])
 => [{}]
2.1.2 :006 > Geocoder::Lookup::Test.add_stub("3", nil)
 => nil
2.1.2 :008 > Geocoder.search("1")
 => []
2.1.2 :009 > Geocoder.search("2")
 => [#<Geocoder::Result::Test:0x007fe54cafac60 @data={}, @cache_hit=nil>]
2.1.2 :010 > Geocoder.search("3")
NoMethodError: undefined method `map' for nil:NilClass
[...]

PR #921, opened 2015-10-02
Test that stubbing with an invaid address works

TrangPham:
for #903, even though there was no issue this time, we should add a test to ensure no issue will be introduced in the future.


alexreisner:
Excellent, thanks!

PR #924, opened 2015-10-09
Remove .near and #nearbys for MongoDB models.

CarlitoBriganteNT:
Geocoding (fetching and saving coordinates) for Mongo-backed models is
still supported, but querying should be done through MongoDB's built-in
commands.


alexreisner:
Great--thanks! I would love to merge this into `drop_mongo_near` but because that branch won't be merged until 1.3.0 you can also remove the deprecation warning...

alexreisner:
Also/separately, I think we can remove the `bson_ext` gem from the various Gemfiles.

PR #932, opened 2015-10-23
Add Mapbox lookup service

AndyDangerous:
Adds lookup using [mapbox](https://www.mapbox.com/developers/api/geocoding/)


AndyDangerous:
It looks like there may be an issue with the regex [here](https://github.com/AndyDangerous/geocoder/blob/mapbox/lib/geocoder/lookups/mapbox.rb#L21). I'm not seeing this locally, but if I can reproduce it then I will fix it.

alexreisner:
Great! Thanks for fixing. Will review this as soon as I can.

PR #944, opened 2015-11-11
Implement Google Places API.

nifarius:


alexreisner:
Thanks, but because Google Places doesn't return latitude and longitude coordinates it's outside the scope of services Geocoder should connect to.

Also, for future reference, this PR breaks tests, which means--generally speaking--the maintainer will probably not look at it until fixed.

nifarius:
It is strange, because Google Places API returns coordinates.
Here is example of response: https://gist.github.com/nifarius/9f73ff2a3e4ab3746b73

alexreisner:
OK, sorry, I missed the fact that the lookup is inheriting from `Google`.

Here's the thing: new lookups means new code to maintain so we need good reasons to justify it. Can you please add a description to this pull request explaining the benefits (what kind of place names can be searched? maybe link to specific pages in the documentation; also confirm that there's no overlap with Google Places Details). Also the test failures will need to be fixed before we can really evaluate.

PR #946, opened 2015-11-17
Mapbox fixes and dataset option

marcusat:
Three change for the Mapbox geocoder:
- Added a new option `:dataset` for Mapbox to specific datasets other than the default `"mapbox.places"`
- Use [long,lat] rather than [lat,long] with Mapbox API
- Added tests more generally to catch this sort of bug

### DIFF
--- a/README.md
+++ b/README.md
@@ -509,6 +509,7 @@ The [Google Places Details API](https://developers.google.com/places/documentati
 #### Mapbox (`:mapbox`)
 
 * **API key**: required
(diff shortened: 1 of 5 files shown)


alexreisner:
Thank you! Only thing I would suggest, in the future, is to make separate commits for each bugfix and feature (ideally, this should have been at least two commits). Small/clean commits are extra important in open source projects where the commit author may be unavailable in the future to help resolve a conflict or trace the origin of a bug.

PR #949, opened 2015-11-26
Add batch reverse geocoding task

FanaHOVA:
Added a batch reverse geocoding task; works just like the batch geocoding one but checks for null `address` rather than lat and lon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,6 +106,10 @@ If you have just added geocoding to an existing application with a lot of object
 
     rake geocode:all CLASS=YourModel
 
+If you need reverse geocoding instead, call the task with REVERSE=true:
+
+    rake geocode:all CLASS=YourModel REVERSE=true
+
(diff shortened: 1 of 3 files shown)


alexreisner:
Thanks for this. However, a few things:

* There is too much code duplication. A new .rake file is not needed and since the method does basically the same thing as the one that does forward geocoding, that code should be reused.
* Your `:not_reverse_geocoded` scope hard codes the `address` column, which is not what everyone will be using to store addresses. Instead, look in the config to get the attribute name (like the `:not_geocoded` scope).

FanaHOVA:
@alexreisner Good points, I threw this together last night to use it in a project of mine, I'll add a flag like `--reverse` instead. 

Re: hardocoding column, I tried using `geocoder_options[:address]` but it returned null, even though it looks like the correct choice looking at the source code.

alexreisner:
I believe it's `:fetched_address` (see: https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/models/active_record.rb#L31)

FanaHOVA:
Should look much better now. Added it to the README as well.

PR #954, opened 2015-12-05
Create a GeoIP2 object for reuse

oogali:
Is there a particular reason why the GeoIP2 lookup is treated as a singleton?

On every call, you're:
- opening the MaxMind database
- doing a lookup
- closing the MaxMind database

This PR allows the caller to reuse the same database handle.

Tests seem to be happy.

image


alexreisner:
Excellent, thanks!

PR #956, opened 2015-12-09
update readme to state how to use ':order => false'

ZebThan:


alexreisner:
This is fine, but I don't think there's any need to mention that passing incorrect method arguments will cause an error. If you can remove that comment I'll merge.

PR #957, opened 2015-12-09
Retrieve city when esri query returns a city POI

roybotnik:
When querying for a city directly (e.g. Washington DC), the current esri result object returns a blank city. 

I tested a wide range of cities and determined that the city attribute is blank for esri results with the types 'City', 'State Capital', and 'National Capital'. However the PlaceName attribute contains the correct city name. I believe this is because the city itself is coming back as a POI, so there's not much point in them duplicating the city value.

This change returns the city by looking at the PlaceName attribute if the result is a city POI. I included tests with the three city location types I was able to determine. I couldn't find a list anywhere in esri's docs, but I queried for a lot of cities and only found these three types.

Let me know if you have any feedback, thanks!

### DIFF
--- a/lib/geocoder/results/esri.rb
+++ b/lib/geocoder/results/esri.rb
@@ -9,7 +9,11 @@ module Geocoder::Result
     end
 
     def city
-      attributes['City']
+      if !reverse_geocode? && is_city?
+        attributes['PlaceName']
+      else
+        attributes['City']
+      end
     end
 
     def state_code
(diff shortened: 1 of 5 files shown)


alexreisner:
Thanks for this, for the sample responses, and the good explanation. Just two things:

1. What do you think about adding an `else` in the `city` method and getting rid of the 'return` statement? I think that would more clearly show the underlying logic (it's either a city or it's not).
2. What do you think about the method name `is_city?` instead of `result_is_city?`. The object is a result, so I don't think we need to reiterate that in the method name.

roybotnik:
Sure thing, thanks for the feedback. I'll update the code with those suggestions.

roybotnik:
OK, updated with the new method name and if/else.

alexreisner:
Looks great! Thanks!

PR #958, opened 2015-12-12
Refactoring Ovi lookup class

davydovanton:


alexreisner:
Hmmm...

* If we freeze the string/name of the API here we should do it for all lookups, and that should be done in a separate pull request where we can discuss the pros and cons of doing so (I'm not sure it's a good idea). Also note that this is not technically refactoring since you're changing the method's return value.
* The dashed like after `private` is a convention in Geocoder and I'd like to keep it (improves readability, I think).
* In `query_url_params` you've changed the order in which hashes are merged. As of today this doesn't matter because they don't share any keys, but this could change, and allowing the shared options to be overwritten by reverse/forward-specific options would become necessary.
* In `api_key` and `api_code` your changes look OK but I don't love the variable name `api`. It should probably be something more specific like `key`, right? (I'm not saying `a` was any better, btw.)

In the end, I really like what you've done on `query_url` and `results`. Can we reduce this PR to just those changes? (I'd be fine keeping `api_key` and `api_code` as well, with the variable name change.)

davydovanton:
sorry, but I don't have a time for this PR

PR #964, opened 2015-12-31
Allow search params to be defined via geocoded_by/reverse_geocoded_by

roybotnik:
This PR allows :params to be specified with geocoded_by and reverse_geocoded_by. They are then passed to Geocoder.search when geocode is called on the model.

As far as I could tell, there was no way to pass `:params => {...}` to Geocoder.search unless it's called manually and I had a situation that kind of needed them. In my particular case, I need to pass `:params => {:category => 'City'}`, which is a thing ESRI supports.

Not sure if this is the right way to go about this. It would be nice to have a test that verifies the param made its way to Geocoder.search, but didn't see a good way to go about it. Let me know what you think, thanks!

### DIFF
--- a/lib/geocoder/models/active_record.rb
+++ b/lib/geocoder/models/active_record.rb
@@ -18,7 +18,8 @@ module Geocoder
           :units         => options[:units],
           :method        => options[:method],
           :lookup        => options[:lookup],
-          :language      => options[:language]
+          :language      => options[:language],
+          :params        => options[:params]
         )
       end
 
@@ -35,7 +36,8 @@ module Geocoder
           :units           => options[:units],
           :method          => options[:method],
           :lookup          => options[:lookup],
-          :language        => options[:language]
+          :language        => options[:language],
+          :params          => options[:params]
         )
       end
 
--- a/lib/geocoder/models/mongo_base.rb
+++ b/lib/geocoder/models/mongo_base.rb
@@ -19,7 +19,8 @@ module Geocoder
           :method        => options[:method],
           :skip_index    => options[:skip_index] || false,
           :lookup        => options[:lookup],
-          :language      => options[:language]
+          :language      => options[:language],
+          :params        => options[:params]
         )
       end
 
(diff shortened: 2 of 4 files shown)


roybotnik:
Another option here would be to add another way to define geocoder options rather than adding them as a geocoded_by arguments. I'd be happy to try that since it would make it easier to expand without gunking up those methods.

TrangPham:
Hi @roybotnik,

Thanks for the pull request. 
There is a test for params making being added to the url via search already: https://github.com/alexreisner/geocoder/blob/master/test/unit/lookup_test.rb#L25
That should be enough since each part of the process is tested.

Putting the :params in geocoder options only would not allow different params for different geocoded models. We should allow both, since the geocode_by (local options) would override the global configurations: https://github.com/alexreisner/geocoder/blob/master/test/unit/configuration_test.rb#L47
Could you add :params into the geocoder options also and add new tests for this? Thanks!

alexreisner:
@TrangPham I agree with that, but we may also need to consider #351. If (some day) we have multiple lookups configured, passing `:params` at the model level is tricky because all lookups accept different params. It's for this reason that I've tried to limit the model-configurable options to globally-applicable things. Does that make sense?

The way things work currently, this would be fine, but I'm a little hesitant to introduce options that might be deprecated in the near future. Thoughts?

TrangPham:
@alexreisner That is true. I was worried that for some of the lookups would have errors since they don't take a :params (For example FreeGeoip). 

The lookups actually get passed the Geocoder::Query object and handle building the query themselves. See: https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookups/esri.rb#L34
Which the other params get added in base lookup: https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookups/base.rb#L133#L141

So for lookups that don't take the :params, they don't use: `url_query_string()` and the :params just get ignored.

@roybotnik Could you also add a `test_query_url_does_not_contains_values_in_params_hash` (similar to https://github.com/alexreisner/geocoder/blob/master/test/unit/lookup_test.rb#L25) Which checks that for lookups that don't take :params, assert_match query url when :param = {} and query url when :param = {:something => "something"}? Thanks

roybotnik:
I had some other, more flexible ideas for this so perhaps I'll give that a shot. Thanks for all of the feedback. I'll have some updates for you soon.

alexreisner:
Closing, because this is old and I'm still not convinced it's a good idea (for reasons above).

PR #967, opened 2016-01-12
Google geocoder results - expose partial_match and place_id fields

rpocklin:
Pretty self-explanatory.  It's great you can access it via `data` but these are very useful fields to expose.


TrangPham:
Thanks for the PR!

Issue #898, opened 2015-08-21
Option to require only needed lookups


dgilperez:
Hi!

I was playing with [derailed_benchmarks](https://github.com/schneems/derailed_benchmarks) to benchmark the memory usage of my app and I realized *geocoder* is one of the most mem-consuming gems in my bundle. This is the derailed output, cut off at 0.8MiB.

    geocoder: 5.9883 MiB
      geocoder/lookup: 5.1445 MiB
        geocoder/lookups/maxmind: 1.4063 MiB
          csv: 1.2539 MiB

The reason for such a footprint is that geocoder is requiring any possible lookup method, even if the app will not use it. I was thinking in a way to require only the lookup methods I was actually going to use, and I found they are hardcoded in [Geocoder::Lookup](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookup.rb). 

Do you think reopening that module and changing `street_services` and `ip_services` is the way to go? Any better idea? Will it actually work? If so, would you accept a PR that adds this feature as a configuration option?

Thanks!

alexreisner:
Hmm...I just checked out derailed_benchmarks and I get the following (similar on multiple projects):

    geocoder: 1.3203 MiB
      geocoder/lookup: 1.2578 MiB
        geocoder/lookups/maxmind: 0.3828 MiB
          csv: 0.332 MiB

Nevertheless, thanks for this, and I am interested in keeping Geocoder's memory use low. It used to be that only the configured lookups would be loaded, but for performance and usability reasons that was changed a couple of years ago. If you can describe your solution in a little more detail (what do you mean by "reopening" Geocoder::Lookup?) I'm open to discussion, but I don't want to lose the performance and usability benefits of loading all the lookups on init. Keep in mind that lookups may not all be specified at load time but selected and configured while the app is running.

dgilperez:
Thanks for the fast response. Actually, I'm seeing varying outputs for derailed_benchmarks depending on the CUT_OFF value, the project or at each execution of the command on the same project ... for the latter, I see values ranging from 6MiB to 2.8MiB. I don't know why is that. 

Anyway, I still think adding the ability to selectively load lookups should improve the memory footprint. I was thinking in a configuration option for Geocoder with the list of the `street_services` and `ip_services` enabled, that would default to the lists now hardcoded [here](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookup.rb#L22). 

[...]

PR #970, opened 2016-01-21
Require only needed lookup

TrangPham:
Connected to #898 

@alexreisner The reason you've required all the lookups seemed to have been for the Test lookup. So I required that by default and then required the other lookups as needed.

For the test_helper, the lookups which are monkey patched needed to be required.
An example of one which isn't required in the test_helper is `yahoo` and at minimum the lookups are all tested in LookupTest `test_search_returns_empty_array_when_no_results`.

### DIFF
--- a/lib/geocoder/lookup.rb
+++ b/lib/geocoder/lookup.rb
@@ -1,3 +1,5 @@
+require "geocoder/lookups/test"
+
 module Geocoder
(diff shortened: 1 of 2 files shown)


alexreisner:
So glad to have this functionality back. Thanks!!

PR #971, opened 2016-01-21
Add missing attribute "types" to Geocoder::Result::Test

arisawa:
Add `types`.
via https://github.com/alexreisner/geocoder/pull/565#issuecomment-30743008. I want it, too.


TrangPham:
See #574. Arbitrary attributes on Result::Test are now allowed. 

Is there a particular reason that types should be hardcorded?

arisawa:
Sorry, I'm really confused..
I couldn't make mock "types" in two weeks ago. But I can make it using v1.2.14 now.
Read the code, certainry it was so. :bow:

PR #974, opened 2016-01-23
Fix Geocoder::Calculations.endpoint comment

dkniffin:
I just spent a few hours trying to get this function to work, then realized I had my parameters in the wrong order. I was following the order listed in the comment, rather than the actual method definition! *facepalm* Anyhow, this fixes the comment, for the next person to read it.

### DIFF
--- a/lib/geocoder/calculations.rb
+++ b/lib/geocoder/calculations.rb
@@ -263,7 +263,7 @@ module Geocoder
     end
 
     ##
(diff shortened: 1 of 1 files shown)


alexreisner:
Ha! Sorry about that. Will certainly merge this, but could you add a comma between "distance" and "provides"? (Improves readability.)

dkniffin:
No worries. I should have been looking at the actual code, rather than the comment. I added the comma.

PR #979, opened 2016-02-04
Suppress Geocoder API warning when no results are returned

rehan-work:
Took a shot at fixing this warning, hopefully I'm doing this right. The Geocoder API warning occurs in the test_no_results unit test, where no results being returned is the expected behaviour. Added a warning suppression line based on prior example.

### DIFF
--- a/test/unit/lookups/geoportail_lu_test.rb
+++ b/test/unit/lookups/geoportail_lu_test.rb
@@ -52,7 +52,7 @@ class GeoportailLuTest < GeocoderTestCase
   end
 
   def test_no_results
-    results = Geocoder.search('no results')
+    results = Geocoder.search('')
     assert_equal 0, results.length
   end
 


alexreisner:
Thanks! This is the correct way to suppress the warning, but issuing a warning for a simple zero-results response doesn't seem like the right behavior (no other lookup does it, and getting no results does not necessarily mean an error occurred). I think the next step is to examine why the lookup is raising an error, and whether it's doing so appropriately (and if it is, let's use a different query to test for no results, because we shouldn't need to suppress warnings in tests unless we're testing something for which we expect a warning to occur).

TrangPham:
Awesome thanks!

Note: other lookup tests that use `Geocoder.search('no results')` worked because there is either a test fixture for the lookup (ie test/fixtures/lookup_name_no_results) or test_helper.rb overrides the result method of the lookup and sets `return [] if query = 'no results'`

PR #980, opened 2016-02-12
Add viewport method to results that provide one

lime:
Several providers return bounding box coordinates for the recommended area to display on a map for a particular query. It thought would be useful to expose this in the same format as bounding boxes are usually passed around in Geocoder.

Feedback would be welcome on naming and format. I see both `[[sw_lat, sw_lon], [ne_lat, ne_lon]]` and `[sw_lat, sw_lon, ne_lat, ne_lon]` being used, but I thought the former might be more clear.

The same method could probably be added to a lot of other providers as well. If you think this is a good idea, I could skim through the fixture data and add the ones I can verify by myself. :)

### DIFF
--- a/lib/geocoder/results/bing.rb
+++ b/lib/geocoder/results/bing.rb
@@ -35,6 +35,10 @@ module Geocoder::Result
       @data['address']
     end
 
+    def viewport
+      @data['bbox']
+    end
+
     def self.response_attributes
       %w[bbox name confidence entityType]
     end
--- a/lib/geocoder/results/esri.rb
+++ b/lib/geocoder/results/esri.rb
@@ -46,6 +46,14 @@ module Geocoder::Result
       [geometry["y"], geometry["x"]]
     end
 
+    def viewport
+      north = attributes['Ymax']
+      south = attributes['Ymin']
+      east = attributes['Xmax']
+      west = attributes['Xmin']
+      [south, west, north, east]
+    end
+
     private
 
     def attributes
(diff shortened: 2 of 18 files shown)


alexreisner:
Thanks! This seems like a useful feature and I'd love to see it on other Result types as well. The only change I'd suggest is that I think the method should be called `viewport`. `bounding_box` usually refers to boundaries for searching, while this is the parameters for displaying a search result.

lime:
I'm very much ok with `viewport`. That was my initial thought, but I felt like maybe I was just following Google's phrasing too much. :)

I'll note though, that this viewport is very much suitable for use as search boundaries. That's actually the use case I had in mind:
```ruby
geocode_results = Geocoder.search('Sweden')
if geocode_results.empty?
  Location.none 
else
  Location.within_bounding_box(geocode_results.first.bounding_box)
end
```

alexreisner:
Haha! Good point. I still think `viewport` is best, though.

lime:
:+1:

lime:
Looking at the code again, I think I should probably use the flattened format `[55.3, 10.9, 69.0, 24.1]` rather than the coordinate-pair format `[[55.3, 10.9], [69.0, 24.1]]`.

It's more consistent with `Calculations.bounding_box`. It also kind of feels more elegant (4 borders vs. 2 corner points).

I'll change the current code to that, and then try to add some more providers.

alexreisner:
That all sounds good to me! Thanks!

lime:
That should be all the viewport responses I could find. I used the fixtures directory as a basis, and tried to verify the format from the online docs of each service.

This is ready to review and merge as far as I am concerned!

alexreisner:
Looks great--thanks for this!

PR #984, opened 2016-02-18
add md5 cache key option to prevent too long cache key

mgraulich:
Similar work was done back in August 2014 but never merged. If you set cache_md5_name: true in the configuration, the md5 value of the URL key will be generated to prevent the file name too long error. 

What do you think?

### DIFF
--- a/lib/generators/geocoder/config/templates/initializer.rb
+++ b/lib/generators/geocoder/config/templates/initializer.rb
@@ -9,6 +9,7 @@ Geocoder.configure(
   # api_key: nil,               # API key for geocoding service
   # cache: nil,                 # cache object (must respond to #[], #[]=, and #keys)
   # cache_prefix: 'geocoder:',  # prefix (string) to use for all cache keys
+  # cache_md5_name: false,      # use MD5 cache name for url? (prevents file name too long error)
 
   # Exceptions that should not be rescued by default
   # (if you want to implement custom error handling);
--- a/lib/geocoder/configuration.rb
+++ b/lib/geocoder/configuration.rb
@@ -53,6 +53,7 @@ module Geocoder
       :api_key,
       :cache,
       :cache_prefix,
+      :cache_md5_name,
       :always_raise,
       :units,
       :distances,
@@ -99,6 +100,7 @@ module Geocoder
       @data[:api_key]      = nil         # API key for geocoding service
       @data[:cache]        = nil         # cache object (must respond to #[], #[]=, and #keys)
       @data[:cache_prefix] = "geocoder:" # prefix (string) to use for all cache keys
+      @data[:cache_md5_name] = false     # use MD5 cache name for url? (prevents file name too long error)
       @data[:basic_auth]   = {}          # user and password for basic auth ({:user => "user", :password => "password"})
       @data[:logger]       = :kernel     # :kernel or Logger instance
       @data[:kernel_logger_level] = ::Logger::WARN # log level, if kernel logger is used
--- a/lib/geocoder/lookups/base.rb
+++ b/lib/geocoder/lookups/base.rb
@@ -218,7 +218,11 @@ module Geocoder
       # The result might or might not be cached.
       #
       def fetch_raw_data(query)
-        key = cache_key(query)
+        if configuration.cache_md5_name
+          key = Digest::MD5.hexdigest(cache_key(query))
+        else
+          key = cache_key(query)
+        end
         if cache and body = cache[key]
           @cache_hit = true
         else


alexreisner:
Thanks for this. For reference, the similar PR is #700.

Let's talk a little about the motivation for this. What are you using for a cache?

mgraulich:
I'm using Rails.cache at the moment but might switch it over to AWS elasticache. The main motivation behind it is to cut down on my IP geocode requests. I currently get the filename too long error when i attempt to cache with my provider.

alexreisner:
I guess what I'm confused about is why you're getting a filename too long error. Why is a file being created? (What is the mechanism behind `Rails.cache` on your app?) Could you post the exact error?

mgraulich:
Got it. Here is the error i get when i disable md5 caching.

2.1.0 :002 > Geocoder.search("162.234.221.46", ip_lookup: :telize)
Errno::ENAMETOOLONG: File name too long @ rb_sysopen - /path/to//tmp/cache/**geocoder%3Ahttps%3A%2F%2Ftelize-v1.p.mashape.com%2Fgeoip%2F162.236.225.100%3Fmashape-key%3D2mOc5vHKI1mshItVb4p2LzsYMjHxp1QrXWkjsnkV60XLDbOO9r20160218-3314-191zrf8**

alexreisner:
OK, I'm guessing you're using ActiveSupport::Cache::FileStore. An alternative way to handle this would be to detect whether `Rails.cache` is defined and whether its return value is an instance of `ActiveSupport::Cache::FileStore`, and to hash they keys if so (rather than using configuration). Problems with this technique include: (1) hash keys dependent on cache mechanism, which seems a little dumb, and (2) we won't successfully detect other filesystem-based caching mechanisms.

I'm thinking out loud here because I don't like the idea of adding configuration for this. It feels too heavy-handed. It's something the app should work out for itself. If it wouldn't break people's existing caches to rename keys in all cases, I'd probably say we should go ahead and do that. However, it will, and that is not an acceptable situation.

What if we were to devise a solution that uses hashed keys but will also work with existing (non-hashed) keys? For example, all new URLs would be stored under hashed keys, but lookups would check for content at both the raw and hashed URLs. The downside of this is that we double the number of cache lookups required in many (if not most) cases. Benefit is that it's largely invisible and doesn't break existing apps.

Any thoughts on this? (cc @TrangPham)

TrangPham:
My opinion is that all caching should use the same hash key mechanism. I also agree that no configuration should be needed. 

Backwards compatibility is important but causing all caching to be slower isn't ideal. If we go with using the raw as a fallback cache lookup then let's make the fallback cache using the new mechanism and clean up/remove the raw key cache. this way eventually the cache will slowly convert itself overtime to being all md5 keys. 

How does that sound?

lime:
Would it be enough to provide an example implementation of a hashing cache? We already provide [three example caches](https://github.com/alexreisner/geocoder/tree/v1.3.0/examples) for auto-expiration and bypass.

I understand that many typical users might be using Rails, and `ActiveSupport::Cache::FileStore` at least in development. To me however, this sounds like a bug in ActiveSupport rather than something we should have to take into account in Geocoder.

alexreisner:
@lime: good idea! I'd be happy to merge an example, and at some point maybe we'll revisit the solution discussed above. At the moment I'm finding it hard to pull the trigger on something that will slow down all caching just to support one (relatively uncommon) caching method.

lime:
I had a look at how `FileStore` already tries to prevent the problem of too long file names, thought you might be interested: [active_support/cache/file_store.rb @ 4-2-stable](https://github.com/rails/rails/blob/4-2-stable/activesupport/lib/active_support/cache/file_store.rb#L119-L138). When a path is deemed too long for the filesystem, they specifically use an MD5 hash of the key instead. 

The assumption seems to be a max filename length of 255 (and max full path of 1024), which applies for many filesystems, but not all. For example, in https://github.com/rails/rails/pull/4911 it is mentioned that ecryptfs would need a significantly lower limit.

There have been some changes for the upcoming version 5 that affect some of these edge cases, e.g. https://github.com/rails/rails/pull/19889. Anyone getting `Errno::ENAMETOOLONG` might want to to give the beta a try just to see if it will have any impact on the issue. If not, maybe a bug report to ActiveSupport is in order?

As for Geocoder, I think this supports the reasoning that we should not be trying to protect the cache from long keys. Even the example hashing cache now feels a bit like a cover-up for upstream issues, but if you still want it I'm happy to put one together.

alexreisner:
@lime thanks for the links and the great discussion! I believe you're right about this being ActiveSupport's problem to solve (and glad to see it's happening soon). Agreed: let's leave this one alone.

PR #985, opened 2016-02-19
Remove Yahoo BOSS support.

galiat:
The Yahoo BOSS Placefinder API is being discontinued.
https://developer.yahoo.com/boss/placefinder/

I'm not sure how this projects prefers to depreciate discontinued APIs. If this is overkill, let me know & I'll just update the readme. The API shuts off March 31, 2016.


alexreisner:
This is good, thanks! Will merge after the shutdown. Until then, though, I think it would be nice to have a deprecation warning stating that the `:yahoo` service will be discontinued on March 31, 2016. I'm thinking the message should appear whenever someone configures `:yahoo` as their lookup and also each time the lookup is used (want to be sure people know that their apps are about to break). Feel free to take a shot at that (separate PR) too if you like.

galiat:
cool. will do

PR #986, opened 2016-02-19
Warn that the Yahoo BOSS Placefinder API is being discontinued.

galiat:
https://developer.yahoo.com/boss/placefinder/

the Yahoo lookup will be fully removed via:
https://github.com/alexreisner/geocoder/pull/985

@alexreisner - I'm not sure whether a warning in `Lookup.get` or in `Lookup::Yahoo.results` makes more sense. Let me know what you think & I will remove one of the warnings.


alexreisner:
Cool. I'd remove the warning from `Lookup.get`. Also better to use the word "discontinued" in the README (rather than deprecated). Other than that, it looks ready to merge. Thanks!

galiat:
ok, done. thanks for the fast feedback!

alexreisner:
Thanks for the quick work! Merged.

Issue #976, opened 2016-01-31
Fix Warnings During Tests


alexreisner:
I'm seeing the following warnings appear during tests with Ruby 2.3.0:

    lib/geocoder/lookups/base.rb:182: warning: constant Geocoder::Lookup::Base::TimeoutError is deprecated
    lib/geocoder/lookups/base.rb:182: warning: constant Geocoder::Lookup::Base::TimeoutError is deprecated
    test/unit/error_handling_test.rb:44: warning: constant ErrorHandlingTest::TimeoutError is deprecated
    test/unit/error_handling_test.rb:49: warning: constant ErrorHandlingTest::TimeoutError is deprecated
    test/test_helper.rb:126: warning: constant Geocoder::Lookup::Base::TimeoutError is deprecated

And also this warning/error (with multiple Ruby versions):

    Geportail.lu Geocoding API error

TrangPham:
Beginner task: 

1) Isolate the test which causes `Geportail.lu Geocoding API error` to be printed out.
Start by running the tests one-by-one and checking output for tests in https://github.com/alexreisner/geocoder/blob/master/test/unit/lookups/geoportail_lu_test.rb

2) Determine if the error should occur

3) If error should occur then change the silence it in the test (see https://github.com/alexreisner/geocoder/pull/843 as an example)
If error should not occur then fix the test

TrangPham:
Thanks for the fixes @lime and @rehan-work (#979) :dancer:

PR #987, opened 2016-02-21
Replace TimeoutError with Timeout::Error

lime:
`TimeoutError` is deprecated in Ruby 2.3.0, causing some of the warnings seen in #976.

It has been an alias for `Timeout::Error` starting in Ruby 1.8.0, and should be safe to replace. Here's the original commit from 2003: https://github.com/ruby/ruby/commit/9d22a06ea06e9bec5030fd6e2c8481958fcd30d9


alexreisner:
Great! Thanks for this!!

PR #990, opened 2016-02-25
Add supported_protocols to Freegeoip.rb lookup

pwoltman:
I uncovered a small issue when switching from Yahoo BOSS to Google (with API, which requires enabling :use_https).  Because the Freegeoip.rb file is missing a supported_protocols method, it attempts to do an https lookup for IP lookups when enabling :use_https.

I already tested this on my setup and it appears to resolve the issue, with lookups/base.rb detecting that https is not supported for Freegeoip


alexreisner:
Thanks for this!

PR #991, opened 2016-03-01
add maxmind_geoip2 setup to README.md

jguitar:
Yesterday, I had to check out the library and search through code to find how to setup `:maxmind_geoip2` as an IP Address Service.

The main goal was _basic_auth_ params. So I think of can be helpful to have this information on the `README.md`.


TrangPham:
Thanks for this!

PR #992, opened 2016-03-01
add mapbox_params to mapbox lookup

AndriySolonyna:
mapbox expects to get  parameters separately from query (example: https://www.mapbox.com/developers/api/geocoding/#country). Сurrently it is impossible to pass some filtering params to query_url .

### DIFF
--- a/lib/geocoder/lookups/mapbox.rb
+++ b/lib/geocoder/lookups/mapbox.rb
@@ -9,7 +9,7 @@ module Geocoder::Lookup
     end
 
     def query_url(query)
-      "#{protocol}://api.mapbox.com/geocoding/v5/#{dataset}/#{url_query_string(query)}.json?access_token=#{configuration.api_key}"
+      "#{protocol}://api.mapbox.com/geocoding/v5/#{dataset}/#{url_query_string(query)}.json?access_token=#{configuration.api_key}&#{mapbox_params_query}"
     end
 
     private # ---------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


TrangPham:
Hi @AndriySolonyna 

Thanks for this PR! I have a couple of suggestions to make this more robust and in-sync with the other lookups of Geocoder.

For other lookups we actually allow the params to be pass in like this:
```
# with Nominatim:
Geocoder.search("Paris", :params => {:countrycodes => "gb,de,fr,es,us"})
```
Or, to search within a particular region with Google:
```
Geocoder.search("...", :params => {:region => "..."})
```
Not through the configuration.

I'll put some in-line comments for more direction :)

lib/geocoder/lookups/mapbox.rb
```diff
@@ -49,5 +49,13 @@ def sort_relevant_feature(features)
         [feature["relevance"],-features.index(feature)]
       end.reverse
     end
+
+    def mapbox_params
+      configuration[:mapbox_params] || {}
+    end
+
+    def mapbox_params_query
+      hash_to_query(mapbox_params)
```
TrangPham: Change this code to:       
```ruby
       hash_to_query(
          query_url_params(query).reject{ |key,value| value.nil? }
        )
```

alexreisner:
Thanks, @TrangPham. Another approach would be to call the `url_query_string` method to generate the entire query string, as is done here: 

https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookups/geocodio.rb#L13

Note the method `query_url_params` which defines all QS variables. This is how lookups are intended to work.

TrangPham:
True, though mapbox overrides url_query_string currently. So if you go with
the method Alex is suggesting remember to rename the existing
url_query_string method in mapbox. I think that's the cleaner way to write
it.

alexreisner:
Closing due to lack of activity.

PR #995, opened 2016-03-07
Add support for the LatLon.io geocoder service

evanmarks:
Adds support for the geocoding (street services) at [LatLon.io](https://latlon.io)

### DIFF
--- a/lib/geocoder/lookups/latlon.rb
+++ b/lib/geocoder/lookups/latlon.rb
@@ -1,0 +1,59 @@
+require 'geocoder/lookups/base'
+require 'geocoder/results/latlon'
+
+module Geocoder::Lookup
+  class Latlon < Base
+
+    def name
+      "LatLon.io"
+    end
+
+    def required_api_key_parts
+      ["api_key"]
+    end
+
+    def query_url(query)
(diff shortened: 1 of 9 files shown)


lib/geocoder/lookups/latlon.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    def query_url(query)
+      "#{protocol}://latlon.io/api/v1/#{if query.reverse_geocode? then 'reverse_' end}geocode?#{url_query_string(query)}"
+    end
+
+    private # ---------------------------------------------------------------
+
+    def results(query)
+      return [] unless doc = fetch_data(query)
+      if doc['error'].nil?
```
TrangPham: Everything looks great! Thanks for the PR

@alexreisner I can't recall from the top of my head your preferences for control structure style? Do you prefer return if/unless more (one lines) instead of if/else (multi lines)?

alexreisner: @TrangPham are you asking about the use of a `return` statement? Basically, this if/else style looks good to me.

If anything, the statement within the string on line 16 could be simplified to: `#{'reverse_' if query.reverse_geocode?}`

evanmarks:
@TrangPham Thanks for the merge and all the work on this project

PR #899, opened 2015-08-21
Added ipinfo.io provider

coderholic:

PR #997, opened 2016-03-12
Added ipinfo.io provider

akostyuk:
Hi,

This is PR related to #899 that has been closed due to inactivity.
I've added sample JSON responses and now tests pass successfully.

Thanks!


lib/geocoder/lookups/ipinfo_io.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    def name
+      "Ipinfo.io"
+    end
+
+    def query_url(query)
+      "#{protocol}://ipinfo.io/#{query.sanitized_text}/geo"
+    end
+
+    # currently doesn't support HTTPS
```
TrangPham: I think we should not restrict this on our side since paid plans actually allow https.
See http://ipinfo.io/pricing

We can make sure the tests skip https check if they fail due to this reason.

TrangPham:
Thanks @akostyuk for picking this up! 

I looked over the code and everything looks good - just my one comment about https. 
Also I checked over the travis build, which is currently failing for the jruby and rbx builds, unrelated to your changes. You PR will be good to merge after the https change.

TrangPham:
Awesome thank you!!

PR #1000, opened 2016-03-15
add pelias/mapzen support

graphaelli:


README.md
```diff
@@ -578,6 +578,17 @@ The [Google Places Details API](https://developers.google.com/places/documentati
 * **Limitations**: ?
 * **Notes**: You can specify which projection you want to use by setting, for example: `Geocoder.configure(:esri => {:outSR => 102100})`.
 
+#### Mapzen/Pelias (`:pelias`)
```
TrangPham: Could you split this documentation into two, one for Mapzen and another for Pelias?
And add short sentences to describe each? eg.
"Mapzen is geocoder powered by open data and uses elastic search"
"Pelias is installed locally and uses the Mapzen API"

`:pelias` should be renamed `:pelias_local` to match the other local lookups.

TrangPham:
Thanks for this PR :) 

The code is good. Please make the documentation in readme more clear. Thanks :dancer:

graphaelli:
Documentation updated per your instructions.  pelias is not strictly local (definitely not local file based).  It's just a self-hosted version of mapzen's service, so we haven't renamed it with `_local`

TrangPham:
Ah makes sense! and awesome thanks for the documentation update. :D

PR #1003, opened 2016-03-18
Added scope :near! for strict results

danielpclark:
Provide strict scope :near! for ActiveRecord::Base instance methods

**Current method implementation with invalid location input.**
```ruby
Address.near("New York, VA").to_sql
# => "SELECT addresses.*, NULL::text AS distance, NULL::text AS bearing FROM \"addresses\"  WHERE (false)" 
```
**New method with invalid location input.**
```ruby
Address.near!("New York, VA").to_sql
#Geocoder::InvalidLocation: The provided location didn't return as valid

```

### DIFF
--- a/lib/geocoder/exceptions.rb
+++ b/lib/geocoder/exceptions.rb
@@ -25,6 +25,9 @@ module Geocoder
   class InvalidRequest < Error
   end
 
+  class InvalidLocation < Error
+  end
+
   class InvalidApiKey < Error
   end
 
--- a/lib/geocoder/stores/active_record.rb
+++ b/lib/geocoder/stores/active_record.rb
@@ -55,6 +55,20 @@ module Geocoder::Store
         }
 
         ##
+        # Strict version of scope :near
+        #
+        scope :near!, lambda{ |location, *args|
+          point = Geocoder::Calculations.extract_coordinates(location)
+          raise Geocoder::InvalidLocation, "The provided location didn't return as valid" \
+            and return if [[Geocoder::Calculations::NAN,Geocoder::Calculations::NAN], nil].
+            include? point
+
+          options = near_scope_options(*Array(point), *args)
+          select(options[:select]).where(options[:conditions]).
+            order(options[:order])
+        }
+
+        ##
         # Find all objects within the area of a given bounding box.
         # Bounds must be an array of locations specifying the southwest
         # corner followed by the northeast corner of the box


danielpclark:
If patch #1004 gets accepted I can modify this to not check for `nil` as well.

danielpclark:
Since all good pull requests require tests - here's mine:

```ruby
gem 'activerecord', '3.2.22.2'
require 'active_record'
require 'minitest/autorun'
require 'logger'

#Ensure backward compatibility with Minitest 4
Minitest::Test = MiniTest::Unit::TestCase unless defined?(Minitest::Test)

# This connection will do for database-independent bug reports.
ActiveRecord::Base.establish_connection(adapter: 'sqlite3', database: ':memory:')
ActiveRecord::Base.logger = Logger.new(STDOUT)

ActiveRecord::Schema.define do
  create_table :addresses, force: true  do |t|
    t.text     "city"
    t.text     "state"
    t.text     "postal_code"
    t.float    "latitude"
    t.float    "longitude"
  end
end

module Geocoder
  module Railtie; end
end
require 'geocoder'
ActiveRecord::Base.extend(Geocoder::Model::ActiveRecord)

class Address < ActiveRecord::Base
  geocoded_by :string_of_address
  after_validation :geocode 

  def string_of_address
    [[city,state].join(","), postal_code].join(" ")
  end
end

class ScopeTest < Minitest::Test
  def test_this_passes
    assert Address.send(:near!, "New York, NY").to_sql
  end

  def test_this_does_not_pass
    assert_raises Geocoder::InvalidLocation do
      Address.send(:near!, [1,1,1,1,1]).to_sql
    end
  end
end
```

And the results are:

```
-- create_table(:addresses, {:force=>true})
D, [2016-03-18T17:21:33.724319 #2839] DEBUG -- :    (0.7ms)  select sqlite_version(*)
D, [2016-03-18T17:21:33.725098 #2839] DEBUG -- :    (0.3ms)  CREATE TABLE "addresses" ("id" INTEGER PRIMARY KEY AUTOINCREMENT NOT NULL, "city" text, "state" text, "postal_code" text, "latitude" float, "longitude" float) 
   -> 0.0070s
Run options: --seed 61835

# Running tests:

..

Finished tests in 0.117916s, 16.9612 tests/s, 16.9612 assertions/s.

2 tests, 2 assertions, 0 failures, 0 errors, 0 skips
```

And to test them you need to have this in the Gemfile

```ruby
gem 'activerecord', '3.2.22.2'
gem 'minitest'
```

alexreisner:
I understand your motivation here, but I don't think this is a good idea for at least a couple of reasons:

1. Elsewhere in the gem, an unsuccessful geocode is not considered an error-worthy condition. If we want to change that, we should change it everywhere (and introduce other `!` methods that raise exceptions in similar conditions, like `Geocoder.search!`), and I don't think the use cases warrant that kind of change.
2. This method is only relevant to the least common (and least recommended) way of calling `near`, which is to pass an un-geocoded string as a center point. I don't like the idea of introducing a method that will encourage a style of usage that's not really recommended (because results of the API call are lost after the query is run).

Sorry!

PR #1004, opened 2016-03-18
Remove chance of nil return in Calculations.extract_coordinates

danielpclark:
This is one place in this method that may return `nil` in `Geocoder::Calculations.extract_coordinates`.

```ruby
Geocoder.coordinates("")
# => nil
```

### DIFF
--- a/lib/geocoder/calculations.rb
+++ b/lib/geocoder/calculations.rb
@@ -416,7 +416,8 @@ module Geocoder
           end
         end
       when String
-        point = Geocoder.coordinates(point) and return point
+        point = Geocoder.coordinates(point)
+        return point ? point : [ NAN, NAN ]
       else
         if point.respond_to?(:to_coordinates)
           if Array === array = point.to_coordinates


alexreisner:
Thanks for this, however I don't totally understand the motivation here. It's pretty standard throughout the gem to return `nil` (instead of `[NAN, NAN]`) when geocoding returns no results. This breaks that convention, and it will break any code in which people are checking for `nil`.

danielpclark:
Perhaps you didn't look at the method.  The last line of the method is `[ NAN, NAN ]`.  To me this clearly defined the expected output if all else failed.  Since this method was designed to return this I believe that the one case where it's a String and calls Geocoder.coordinates and it may return `nil`, was an oversight, as it didn't account for the illegitimate String case.

I did not design the `[ NAN, NAN ]` as a results, it's already in the design.  I took the time to understand the design intent behind the method before applying this patch.  Personally I'd rather have `nil` in both cases of this method.  But I'm not one to go against the intent of a projects design.

alexreisner:
Ah, yes, you're right. However, on closer inspection, that is already the method's behavior--it doesn't return `nil` when geocoding of the string fails (the `return` is not evaluated and it gets to the end of the method where it returns `[NAN, NAN]`). Explanation of why the array is returned instead of `nil` is here: d81b08b2ca43732601ddd3883db06a6977afb51a.

danielpclark:
Oh yeah.  My bad.  When `nil` gets returned and then calls the `and` before `return` it stops before returning because `nil` is false so the `and` already met a failed criteria.  My bad.

PR #1007, opened 2016-03-24
Support for custom telize endpoint

jwaldrip:
Telize offers an opensource version that can be run from a custom host. This should add support for it.


TrangPham:
Thanks for this PR :)

Could you add a note to the readme under Telize lookup?

TrangPham:
Created a task for the readme update so this could be merged.

PR #1011, opened 2016-03-24
Make geocoded? return false with single coordinate

### LINKED ISSUE #1010: geocoded? returns true even if a single coordinate is nil
It's an edge case, but if coordinates happen to be editable by users or something, you could end up with a single coordinate being nil.  In this case, I believe the `geocoded?` method of a model should return false, but it returns true, because it's only checking that there are > 0 coordinates present.  It should check for exactly 2.

mltsy:
fixes #1010

### DIFF
--- a/lib/geocoder/stores/base.rb
+++ b/lib/geocoder/stores/base.rb
@@ -6,7 +6,7 @@ module Geocoder
       # Is this object geocoded? (Does it have latitude and longitude?)
       #
       def geocoded?
-        to_coordinates.compact.size > 0
+        to_coordinates.compact.size == 2
       end
 
       ##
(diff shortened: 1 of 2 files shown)


mltsy:
Huh - tests failed on CI, but I think it's something unrelated...?
```
An error occurred while installing mime-types-data (3.2016.0221), and Bundler cannot continue.
Make sure that `gem install mime-types-data -v '3.2016.0221'` succeeds before bundling.
```

alexreisner:
Yeah, that's unrelated. Don't worry about it, the code looks good. Will merge into a new branch or at a later date into master. Thanks!

PR #1016, opened 2016-03-30
Enabling ESRI for storing results

erdostom:
- `api_key` param
- `for_storage` param
- token generation
- updating README.md

ESRI needs authentication for paid requests and to avoid rate limiting.

https://developers.arcgis.com/rest/geocode/api-reference/geocoding-free-vs-paid.htm


TrangPham:
Thanks for the PR!

The tests seem to be failing on due to `configuration.for_storage`.

Also I'm not sure that the token regeneration should live within Geocoder. @alexreisner What do you think of this?

alexreisner:
Hmm, yes the token generation may be a little tricky. @erdostom: can multiple ESRI tokens be active at any given moment? (The documentation seems unclear.) If an app is running on multiple servers, each will be generating their own token (each stored in RAM), possibly invalidating the tokens previously generated by other instances. But if multiple tokens are OK then I think token generation is fine to do in this way. @TrangPham did you have other concerns about it?

erdostom:
@alexreisner multiple tokens are OK as long as they are unexpired.

alexreisner:
OK, then I think at the very least the token generation code should be decoupled from the token storing code and made available as a public class method on the ESRI lookup so that if someone does want to share a single token among servers they can use that method for generating it. In that case we should also have a way of setting the token manually, possibly via `Geocoder.configure`. In fact, that's perhaps where the token should always be stored, rather than in an instance variable.

What I'm not sure of is how the token expiration time should be stored. Maybe it also goes into configuration? Maybe the token is an object (`EsriToken`?) that responds to `to_s` and `expires_at`? The latter seems cleaner (token-related functionality could live in that class) but I'm open to discussion on this.

erdostom:
I'm OK with this approach. Where should I create the `EsriToken` object? Do you think this warrants a new directory under `/lib/geocoder/`?

alexreisner:
OK, great. I think `lib/geocoder/esri_token.rb` is fine for now. If there was another, similar type of thing somewhere then I'd suggest we put them in a directory together, but as it stands it's kind of unique and I think it's alright to put it there. Thanks!

TrangPham:
Awesome that's a much cleaner solution I think. I'm good with this plan.

erdostom:
@alexreisner @TrangPham please take a look at the updates and let me know if it's good to go. Thank you

alexreisner:
Thanks! The important parts of this look good, however let's talk about the following:

* Config for `:for_storage` and `:token` are specific to the `:esri` lookup so they don't need to be added to the global configuration hash.
* I think `valid?` is a slightly inaccurate method name (implies some kind of validation is being done on the string). What about something like `active?` or `expired?`?
* You should avoid parsing JSON twice (happens when there is no error).
* Are there other errors that could occur on token generation besides InvalidApiKey? As is, we may be making an incorrect assumption. Actually, now that I think more about this, we should probably have a timeout on the request, like we have for geocoding requests. Hmmm...it may be a lot of work to really do this properly (need to handle all error conditions that we handle when making geocoding requests), so maybe we just leave it as is for now and come back to it later. But I think at the very least we should display the actual error returned by ESRI (in the JSON).
* We should really have a test or two for this new functionality. Mocking network requests seems out of scope, but can we add a test for the lookup and use an EsriToken we create manually?

lib/geocoder/lookups/esri.rb
```diff
@@ -15,6 +16,24 @@ def query_url(query)
         url_query_string(query)
     end
 
+    def generate_token(expires=1440)
```
TrangPham: Putting this method into the EsriToken class would be cleaner, could you move it over? Thanks

erdostom:
If I take `:for_storage` and `:token` out of the global config hash, my tests error. Where should I place them, esp so that `Geocoder.configure` still has access to :`token`? Thanks

alexreisner:
Try setting those config vars as ESRI-specific, eg: `Geocoder.configure(esri: {token: "..."})`

erdostom:
@alexreisner already doing that, but it seems like i can't access `configuration.token` unless I add it to `OPTIONS` in `configuration.rb`

erdostom:
Please take a look at latest, it should address the issues except the global config hash. Thank you

alexreisner:
Try `configuration[:token]`. Then I'll take a look at everything at once.

erdostom:
That worked. Thanks @alexreisner

test/unit/lookups/esri_test.rb
```diff
@@ -22,12 +22,12 @@ def test_query_for_geocode_with_token_for_storage
     query = Geocoder::Query.new("Bluffton, SC")
     lookup = Geocoder::Lookup.get(:esri)
     res = lookup.query_url(query)
+    Geocoder.configure(esri: {token: nil, for_storage: nil})
```
alexreisner: This should probably go in `teardown` in case the test exits before it gets to this line.

erdostom:
let me know if you need anything else to get this PR merged. thank you

PR #1018, opened 2016-04-03
Add ip api com lookup

piotrgorecki:
Hi
I prepared support for IP-API.com ip lookup service, basing on the ipinfo_io code.
My tests passed but I don't have 100% reliability of code, because I'm able to use it only at localhost.
Btw, this is my first pull request :)


test/unit/lookup_test.rb
```diff
@@ -24,7 +24,7 @@ def test_search_returns_empty_array_when_no_results
 
   def test_query_url_contains_values_in_params_hash
     Geocoder::Lookup.all_services_except_test.each do |l|
-      next if [:freegeoip, :maxmind_local, :telize, :pointpin, :geoip2, :maxmind_geoip2, :mapbox, :ipinfo_io].include? l # does not use query string
+      next if [:freegeoip, :maxmind_local, :telize, :pointpin, :geoip2, :maxmind_geoip2, :mapbox, :ipinfo_io, :ipapi_com].include? l # does not use query string
```
TrangPham: ipapi_com should not be included here as the API does take params.

TrangPham:
@piotrgorecki 

Thanks for your PR! And congrats on your first PR :)

You tests have passed and your code looks generally good. I have a few nitpicks though.

1) Preferably the results class would be cleaner. We don't need to have default responses for the fields.

2) I checked the API and it already handles unknown fields being passed in - so we don't need to be checking them in the query url params.  
3) Most of the other lookups allow the params to also be based on the individual query level. This should also be true for ipapi_com. Such that we could also use `Geocoder.search('74.200.247.59', :params => {:fields => 'ciy,lat,lon,whatever'})`

piotrgorecki:
I fixed 1) and 2). 
Regarding 3), I wrote one test for using parameters directly in search method and it passed, so I didn't change anything else.

TrangPham:
Thanks so much!

PR #1020, opened 2016-04-04
Use a proper domain name in ipapi_com lookup

piotrgorecki:
I omitted that I should use different domain name in case when an api key is in use.
Moreover, when the api key is invalid, the service respond "invalid key" - should I handle this?

My original pull request: [https://github.com/alexreisner/geocoder/pull/1018](https://github.com/alexreisner/geocoder/pull/1018)


TrangPham:
Yes - the invalid key can be handled in the result method of the lookup

piotrgorecki:
I handle it in parse_raw_data, because the service's response is the plain text "invalid key\n", not any json.

TrangPham:
@piotrgorecki 

For other invalid keys in geocoder, we use
```
          raise_error(Geocoder::InvalidApiKey) || Geocoder.log(:warn, "Invalid API key.")
```

Could you use this for the invalid key to make it uniform with the other geocoder lookups? This will help geocoder users catch the error.

TrangPham:
Thanks so much @piotrgorecki 👍 Looks like people have already started using your new addition!

PR #1024, opened 2016-04-12
Add district to google result

defkode:


alexreisner:
Thanks for the PR. However, I've rejected similar PRs for the reason that they end up being misleading. Google's `sublocality` will only be a "district" in certain cities/countries/regions. What's called a district locally in one part of the world might be `administrative_area_level_3` in another. Google has good reasons for using such generic labels in their data. Renaming is not advised.

Issue #1009, opened 2016-03-24
Error with freegeoip: response was not valid JSON


arnlen:
When using the FreeGeoIP provider, I get this error message:

```
Geocoding API's response was not valid JSON: ({"ip":"90.34.122.47","country_code":"FR","country_name":"France","region_code":"O","region_name":"Nord-Pas-de-Calais","city":"Lumbres","zip_code":"62380","time_zone":"Europe/Paris","latitude":50.7069,"longitude":2.1208,"metro_code":0});
```

I guess it's because of the parenthesis.
I've opened this [issue on the freegeoip repository](https://github.com/fiorix/freegeoip/issues/167) to get more info about why the API is returning this payload.

From the Geocoder side: is there something we can do?

alexreisner:
On hold until we see what happens with freegeoip.

profeg:
freegeoip.net is down now.  
Available freegeoip.io

PR #1025, opened 2016-04-12
change freegeoip lookup's host from freegeoip.net to freegeoip.io

lowellmower:
Given 'freegeoip.net' is down for the foreseeable future ([ref here](https://twitter.com/fiorix_/status/715892818758221825)) and a clone has been established at 'freegeoip.net' I propose updating the default IP lookup to the latter. This is also somewhat in response to issue [#1009](https://github.com/alexreisner/geocoder/issues/1009)

### DIFF
--- a/lib/geocoder/lookups/freegeoip.rb
+++ b/lib/geocoder/lookups/freegeoip.rb
@@ -45,7 +45,7 @@ module Geocoder::Lookup
     end
 
     def host
-      configuration[:host] || "freegeoip.net"
+      configuration[:host] || "freegeoip.io"
     end
   end
 end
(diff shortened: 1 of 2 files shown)


lowellmower:
@alexreisner let me know if you see an issue with this or if you have an alternative approach you can think of

TrangPham:
This looks good to me. Tested the json end point and compared it to what's expected from results.rb. Only difference was metrocode (results.rb) verses metro_code (json response). 

Could you update results.rb for this? Thank you :)

lowellmower:
👍

alexreisner:
Also added: 30f332c8b927407b257a702981ce28db07cf4c9c

PR #1031, opened 2016-04-19
adequation to neighborhood method of Results::Google object

rplaurindo:
The ```types``` key doesn't more contains ```neighborhood``` symbol, then the method always returns ```nil``` value, now contains ```sublocality``` and ```sublocality_level_1```.

The ```types``` key doesn't more contains ```neighborhood``` symbol, but now contains ```sublocality``` and ```sublocality_level_1```, for that the ```neighborhood``` method always returns nil value.

### DIFF
--- a/lib/geocoder/results/google.rb
+++ b/lib/geocoder/results/google.rb
@@ -12,7 +12,7 @@ module Geocoder::Result
     end
 
     def neighborhood
-      if neighborhood = address_components_of_type(:neighborhood).first
+      if neighborhood = address_components_of_type(:sublocality).first
         neighborhood['long_name']
       end
     end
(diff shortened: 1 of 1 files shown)


TrangPham:
Hi @rplaurindo 

The google geocoding API states that neighborhood is in fact a type: https://developers.google.com/maps/documentation/geocoding/intro#Types

If you are receiving a nil value it is because that is what the google API is returning for neighborhood. If you have a specific example where the google API is returning neighborhood but geocoder is listing neighborhood as a nil, please share it with us.

Thank you :)

rplaurindo:
@TrangPham hello, sorry for late. Address from Brazil apparently doesn't have neighborhood, but a equivalence can be identified in ```["sublocality_level_1", "sublocality", "political"]```.
You can see through lat long pair ```-21.750663,-41.331070```, or directly by [url](https://maps.googleapis.com/maps/api/geocode/json?latlng=-21.750663,-41.331070).
Maybe a better solution would be:
```ruby
def neighborhood
  if neighborhood = address_components_of_type(:sublocality).first || address_components_of_type(:neighborhood).first
    neighborhood['long_name']
  end
end
```

dukex:
Hey @TrangPham can you validate this issue again? I can open a PR about it using the code @rplaurindo already made

PR #1033, opened 2016-04-28
Add result_type as option for google lookup

hanloong:
Hey guys.

Just wondering if it would be possible have the option to return `result_type` from the google geocoder reverse lookup as seen here.
https://developers.google.com/maps/documentation/geocoding/intro#ReverseGeocoding


alexreisner:
Looks good to me. Thanks!

Issue #1026, opened 2016-04-12
New Default Street Address Lookup


alexreisner:
I think it would be great if Geocoder's default street address lookup was an API powered by open source code and/or data. However, one of Geocoder's primary goals is to be easy to set up and use, especially for the first time. Hence, any alternative to Google should meet the following requirements:
- no API key required
- covers most of the world
- returns helpful error messages (and lookup supports raisable exceptions)

The leading candidate as of today would seem to be Nominatim. This change would break some apps so it will only be made for a minor (not patch) version release.

All thoughts, suggestions/complaints, and proposals welcome.

PR #1035, opened 2016-05-02
Disable SSL v2 and v3 if ssl enabled

### LINKED ISSUE #1034: SSL error - Google
I've been seeing sporadic issues with google's SSL handshake:

`SSL_connect SYSCALL returned=5 errno=0 state=SSLv3 read server session ticket A`

I'm using [this fix](https://github.com/jlhonora/geocoder/commit/530fc558a33400aad4dc905cf912f3162f9c950b) in production:

```
diff --git a/lib/geocoder/lookups/base.rb b/lib/geocoder/lookups/base.rb
index eb39893..956f050 100644
--- a/lib/geocoder/lookups/base.rb
+++ b/lib/geocoder/lookups/base.rb
@@ -274,6 +274,12 @@ module Geocoder
         uri = URI.parse(query_url(query))
         Geocoder.log(:debug, "Geocoder: HTTP request being made for #{uri.to_s}")
         http_client.start(uri.host, uri.port, use_ssl: use_ssl?, open_timeout: configuration.timeout, read_timeout: configuration.timeout) do |client|
+          if use_ssl?
+            client.instance_eval {
+              @ssl_context = OpenSSL::SSL::SSLContext.new
+              @ssl_context.set_params({:options=> OpenSSL::SSL::OP_NO_SSLv2 | OpenSSL::SSL::OP_NO_SSLv3 | OpenSSL::SSL::OP_NO_COMPRESSION})
+            }
+          end
           req = Net::HTTP::Get.new(uri.request_uri, configuration.http_headers)
           if configuration.basic_auth[:user] and configuration.basic_auth[:password]
             req.basic_auth(
```

but I'm not sure if it fixes the problem entirely. Besides, there could still be some lookups where SSLv3 (or lower) is needed.

Anyone else experiencing this?

jlhonora:
Fixes #1034 

About the test, I didn't feel comfortable changing such a crucial piece of the code without at least making a test that passes :) . Basically what I did is rename the `:make_api_request` to `:make_http_api_request`, and then verify the behavior of the latter.

### DIFF
--- a/lib/geocoder/lookups/google.rb
+++ b/lib/geocoder/lookups/google.rb
@@ -30,7 +30,11 @@ module Geocoder::Lookup
     def configure_ssl!(client)
       client.instance_eval {
         @ssl_context = OpenSSL::SSL::SSLContext.new
-        @ssl_context.set_params({:options=> OpenSSL::SSL::OP_NO_SSLv2 | OpenSSL::SSL::OP_NO_SSLv3 | OpenSSL::SSL::OP_NO_COMPRESSION})
+        options = OpenSSL::SSL::OP_NO_SSLv2 | OpenSSL::SSL::OP_NO_SSLv3
+        if OpenSSL::SSL.const_defined?('OP_NO_COMPRESSION')
+          options |= OpenSSL::SSL::OP_NO_COMPRESSION
+        end
+        @ssl_context.set_params({options: options})
       }
     end
 
--- a/lib/geocoder/lookups/google.rb
+++ b/lib/geocoder/lookups/google.rb
@@ -27,6 +27,13 @@ module Geocoder::Lookup
 
     private # ---------------------------------------------------------------
 
+    def configure_ssl!(client)
+      client.instance_eval {
+        @ssl_context = OpenSSL::SSL::SSLContext.new
+        @ssl_context.set_params({:options=> OpenSSL::SSL::OP_NO_SSLv2 | OpenSSL::SSL::OP_NO_SSLv3 | OpenSSL::SSL::OP_NO_COMPRESSION})
+      }
+    end
+
     def valid_response?(response)
       json = parse_json(response.body)
       status = json["status"] if json
--- a/Gemfile
+++ b/Gemfile
@@ -24,6 +24,7 @@ end
 
 group :test do
   gem 'sqlite3', :platform => [:ruby, :mswin, :mingw]
+  gem 'webmock'
 
   platforms :ruby do
     gem 'pg'
--- a/gemfiles/Gemfile.rails3.2
+++ b/gemfiles/Gemfile.rails3.2
@@ -24,6 +24,7 @@ end
 
 group :test do
   gem 'sqlite3', :platform => [:ruby, :mswin, :mingw]
+  gem 'webmock'
 
   platforms :ruby do
     gem 'pg'
--- a/gemfiles/Gemfile.rails4.1
+++ b/gemfiles/Gemfile.rails4.1
@@ -24,6 +24,7 @@ end
 
 group :test do
   gem 'sqlite3', :platform => [:ruby, :mswin, :mingw]
+  gem 'webmock'
 
   platforms :ruby do
     gem 'pg'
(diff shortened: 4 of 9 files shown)


jlhonora:
Ooops, sorry for the failed build, missed the `gemfiles` folder. Is there a way to run the build matrix locally?

TrangPham:
Hi @jlhonora 

The way I have been running the build martix is to push to my own fork, which I have enabled on Jarvis to build. This is pretty simple way to do it, but you can also feel free to keep pushing to your PR and Jarvis will keep rerunning the build martix.

lib/geocoder/lookups/google.rb
```diff
@@ -27,6 +27,13 @@ def query_url(query)
 
     private # ---------------------------------------------------------------
 
+    def configure_ssl!(client)
+      client.instance_eval {
+        @ssl_context = OpenSSL::SSL::SSLContext.new
+        @ssl_context.set_params({:options=> OpenSSL::SSL::OP_NO_SSLv2 | OpenSSL::SSL::OP_NO_SSLv3 | OpenSSL::SSL::OP_NO_COMPRESSION})
```
TrangPham: Why do we need to change the SSL context? Also is there a reason that this is only relevant to google's API?

jlhonora:
Hi @TrangPham , thanks! Will check on that soon and get back to this.

alexreisner:
Ha! I like your testing instincts. However, this test isn't really related to the functionality being added. I just started a new issue about testing (#1040) where we could discuss this further.

Also, as @TrangPham mentioned, could you briefly explain (or, even better, provide a link to an explanation of) why the change is necessary with Google?

jlhonora:
Re: testing; my thoughts exactly. I just wanted to make sure I'm not breaking http requests. Although it doesn't test the change, it did prove useful in the ruby 1.9 build scenario.

Re: TLS @ Google; this is the most official statement I could find: https://developers.google.com/web/updates/2016/04/geolocation-on-secure-contexts-only . It mentions Chrome specifically, but I've seen similar enforcements in their geocoding API. The issue doesn't arise consistently. If you try a request it will (probably) complete successfully:

```
$ curl -I --sslv3 "https://maps.googleapis.com/maps/api/geocode/json?address=Menlo+Park"
HTTP/1.1 200 OK
```

but we have a production workload where the following exception showed up randomly (in about 0.01% of 200k daily requests) when performing `Geocoder.search`:

```
SSL_connect SYSCALL returned=5 errno=0 state=SSLv3 read server session ticket A
```

Perhaps Google has some servers that are already enforcing this? Maybe for the premier/enterprise plan only? Obviously I'm not sure.

Anyhow, as Google's post mentions, SSLv3 is broken and should not be allowed, even more in privacy-sensitive contexts such as location.

alexreisner:
Hmm...I'm not seeing anything about SSL 3.0 in that post, but are you referring to [the vulnerability discovered in Oct 2014](https://en.wikipedia.org/wiki/Transport_Layer_Security#SSL_3.0)?

alexreisner:
Regardless I think this change is okay, and the fact that it's working for you in production is reassuring.

It's also true that the test revealed something useful in Ruby 1.9.3. What do you think about the following changes:

* When the method is aliased in `test_helper.rb`, change the name to something more descriptive (eg: `actual_make_api_request`) and add a code comment describing why it's being done (eg: make method available for direct testing).
* For the test itself, use a more descriptive name (eg: `test_actual_make_api_request_with_https`).
* Also worth noting that if we want to be testing the Google lookup then we should configure `lookup: :google` at the beginning of the test and not assume that Google will always be the default (note #1026). This sort of implies that the test should be moved to `google_test.rb`, although it's probably useful to have in both places (testing the default lookup seems smart).

jlhonora:
Sorry for taking this long, and thanks a lot for the insights.

You're right, no mention of SSL in that post. I applied the changes based on your suggestions, but I'm ok with leaving this out until there's an official announcement with respect to older SSL/TLS versions.

alexreisner:
I spoke to someone at Google who said all connections (on their end) should be using TLS. That doesn't appear to be the case but does it affect anything we're doing here? We're only preventing SSLv3, right?

jlhonora:
With this change we're effectively allowing only TLS 1.0 and up. That leaves the protocol selection to the client-server negotiation. A common setup is that the server config is preferred, and the highest common configuration is chosen.

alexreisner:
OK, thanks for the clarification (and all the good work). Merging now.

PR #1043, opened 2016-05-16
Add warning for redis cache clearing

joelr:
.keys in a Production environment is pretty dangerous, particularly when it's being manipulated in ruby as an array afterwards.

This should probably be reworked to use something like scan_each, eg:

``` ruby
store.scan_each(:match => "geocoder:*") { |key|
  store.del(key)
end
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -868,6 +868,8 @@ By default the prefix is `geocoder:`
 
 If you need to expire cached content:
 
+_Warning: If you are using redis for caching, this method of caching uses .keys to select URLs to expire, with regex processing in ruby to filter geocoder URLs/keys to delete. This in a large production environment can cause Redis to lock, and will most likely cause memory issues._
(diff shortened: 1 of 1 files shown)


alexreisner:
The challenge here is keeping things general enough that the code works with a variety of cache back ends--`scan_each` will only work with Redis. Because it's so popular, though, it might be worth having the `expire` method check whether the cache responds to `scan_each` before calling `urls.each`.

It looks like the only time `keys` is called is on `expire(:all)`. Could you: (1) be more specific about that in the warning, (2) move the warning to the bottom of the Caching section, and (3) capitalize all occurrences of "Redis," for consistency.

Thanks!

alexreisner:
Closing, as this warning seems like the wrong solution. If this is a concern, it should be changed in the code.

Issue #1044, opened 2016-05-16
Add support for batch geocoding


aaronpk:
Some providers (such as Esri) support an API method for batch geocoding. In the case of Esri, very different rate limits apply, so we can use this to geocode a large number of records without worrying about hitting HTTP rate limits.

I plan on using the Esri batch geocoding API, and thought it would be a good thing to add to this gem. Any thoughts on whether a PR adding support for that would be accepted?

alexreisner:
Generally speaking, I'm open to adding this feature. Would it be added through a rake task? It seems like it might be neat if the existing batch geocoding rake task was modified to detect whether the configured lookup supports batch geocoding, and took advantage of the feature if available.

aaronpk:
well I plan on using this without any of the model support, so I will need to either add support for it to `Geocoder.search` or make a new method like `Geocoder.batch`. I do like the idea of the Rake task using batch geocoding if it's configured, although that isn't something I will be using.

TrangPham:
I would prefer it to be Geocoder.batch then adding complexity to .search

alexreisner:
@TrangPham agreed

aaronpk:
I've made some progress on this here: #1046 I'd appreciate a quick 👍 if I'm on the right track with how others imagine this feature working.

henry-p:
@alexreisner I just discovered that - at least for Google Maps -, batch geocoding already works by passing an array of street addresses: `Geocoder.search(["1 abc street, foo-town", "2 xyz stree, bar-town", "..."])`. It works because the method accepts an array, in order to support an array of lat/lon strings (-> a lat/lon-pair). It then concats these strings and makes the query for multiple street addresses in one request. Judging from the comments in the code, this was *not* intentional (?). But we actually appreciate that feature, as it improves our code performance significantly. Is there any way to keep the existing functionality of the code (such that you can continue using the `search` method with an array), but improve the structure such that it becomes apparent that batch searching is possible (e.g. using the `options`)? Hope this doesn't sound like a bunch of nonsense...

alexreisner:
Haha! @henry-p you're right that wasn't intentional, but I'm glad it's working for you. However, I can't reproduce it on my end, and I wasn't aware Google provided that service. Could you post a sample response? (I get a 400 Bad Request when I try it.)

tuckerbuchy:
Any updates on this feature? I noticed that Bing Maps has a batch option too and want to access it. 
[...]

Issue #603, opened 2014-02-19
Sensis support


DanielHeath:
I've implemented sensis geocoder support here:

http://github.com/danielheath/geocoder-sensis

I'd be happier to see it merged into core rather than kept as an extension though.

TrangPham:
I can't seem to find information about the sensis geocoding services via googling - I saw a couple results which could be related. Looks like it's a service which you must host yourself? (based on the code in geocoder-sensis)

It looks like there are people who are using this API though: https://rubygems.org/gems/geocoder-sensis/versions/0.0.4

Adding to beginners tasks:
- Port Sensis lookup over into Geocoder (see #591 as example) 
- Add tests
- Add documentation in read me

alexreisner:
Closing, as this API no longer seems to exist.

PR #1046, opened 2016-05-18
initial draft of batch geocoding

aaronpk:
- Adds new `Geocoder.batch` method which takes an array of input text and uses the provider's batch API to make a single HTTP request to geocode all the inputs.
- Updates readme with a usage example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -949,6 +949,19 @@ You can use Geocoder outside of Rails by calling the `Geocoder.search` method:
 This returns an array of `Geocoder::Result` objects with all data provided by the geocoding service.
 
 
+Batch Geocoding Outside of Rails
+--------------------------------
+
+For providers which have a batch geocoding API, you can use the `Geocoder.batch` method to pass in multiple addresses and get back a list of corresponding geocoded results. This will make a single API request to the provider, which will result in a significantly faster response when geocoding a large number of items.
+
+    results = Geocoder.batch([
+      {input: "309 SW 6th, Portland, OR", id: 2042},
+      {input: "220 NW 8th Ave, Portland, OR", id: 3029}
+    ])
+
+This returns an array of `Geocoder::Result` objects. If you provide an `id` for each input, the result object will have a matching id. Otherwise, the id will be automatically assigned.
+
+
 Testing Apps that Use Geocoder
 ------------------------------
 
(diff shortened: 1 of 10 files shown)


aaronpk:
for #1044

aaronpk:
I've started a draft of how this might work. I'd appreciate a quick review before I go any farther with it. I want to make sure I'm not missing any assumptions. I've only added batch support to the Esri provider, since I'm not sure if other providers have a batch API. The Esri service requires authentication in order to use the batch API, so if you want to run it yourself you'll need to register for a developer account at https://developers.arcgis.com

TrangPham:
It looks good so far! :) 

Could more of the functionality for batch be handled in the base lookup? 
Ideally implementing batch for a lookup should be simple and easy to understand. 
@alexreisner Do you have suggestions?

alexreisner:
I actually like the separation of batch and single requests. In fact, I wonder whether we shouldn't go a step further and use entirely separate lookups for batch requests. Reasons include: (1) combining them is potentially misleading, since only one lookup has batch capability, (2) single and batch queries are used in totally different settings, and (3) there's no reason to expect future batch APIs to be as similar to their single-query counterparts as ESRI's are. Conversely to (3), they could be even more similar than ESRI's, so that argument works both ways.

The lack of a second batch API example is what makes this whole thing a little difficult to think about. I want to choose an architecture here that minimizes the chances of problems in the future. We should really look at how other batch APIs are implemented. I believe [Bing](https://blogs.bing.com/maps/2010/08/31/batch-geocoding-and-batch-reverse-geocoding-with-bing-maps/) and Mapquest provided batch lookups at one time but I'm not sure whether they still do.

As for the code, I have the following thoughts:

* as discussed in #1058, don't add the `use_post` config option (should be automatic); `create_http_request` should take a `verb` argument
* the new class should be called BatchQuery, for clarity
* `BatchQuery` should probably inherit from `Query`, since it has a similar interface
  * we want the `url` method, esp for testing (provided by `Query`)
  * shouldn't copy/paste `lookup` method
  * `reverse_geocode?` should have a real implementation
* remove debugging line (`jj doc`)
* be careful of mixing refactoring in with this PR (I don't love the changes in `Result::ESRI` which make it less explicit when the different hash keys will be used

zhaomengru2015:
any updates on this PR? I am currently working on a  project which is  using sensis service, I am trying to implement a PR about [sensis support](https://github.com/alexreisner/geocoder/issues/603), which will use the 'post' http request.

alexreisner:
Closing this due to lack of activity, and the fact that creating a unified interface for bulk geocoding is quite difficult given the wide range of services offered (see comment above).

PR #1047, opened 2016-05-19
fix passing forStorage=true for esri geocoder when using api key

aaronpk:
Previously, the esri config object was being replaced when setting the token so the `for_storage` option was cleared out. This meant that anyone using the `api_key` (client credentials) method of authenticating as well as trying to set the `for_storage` option was unable to actually pass the parameter to the Esri geocoder, breaking their TOS.

### DIFF
--- a/lib/geocoder/lookups/esri.rb
+++ b/lib/geocoder/lookups/esri.rb
@@ -52,7 +52,14 @@ module Geocoder::Lookup
         configuration[:token].to_s
       elsif configuration.api_key # generate a new token if we have credentials
         token_instance = Geocoder::EsriToken.generate_token(*configuration.api_key)
-        Geocoder.configure(:esri => {:token => token_instance})
+
+        esri_config = {:token => token_instance}
+        if configuration[:for_storage]
+          esri_config[:for_storage] = configuration[:for_storage]
+        end
+        # replaces the entire :esri config section with the new object
+        Geocoder.configure(:esri => esri_config)
+
         token_instance.to_s
       end
     end
--- a/lib/geocoder/lookups/esri.rb
+++ b/lib/geocoder/lookups/esri.rb
@@ -53,12 +53,7 @@ module Geocoder::Lookup
       elsif configuration.api_key # generate a new token if we have credentials
         token_instance = Geocoder::EsriToken.generate_token(*configuration.api_key)
 
-        esri_config = {:token => token_instance}
-        if configuration[:for_storage]
-          esri_config[:for_storage] = configuration[:for_storage]
-        end
-        # replaces the entire :esri config section with the new object
-        Geocoder.configure(:esri => esri_config)
+        Geocoder.configure(:esri => Geocoder.config[:esri].merge({:token => token_instance}))
 
         token_instance.to_s
       end
--- a/test/unit/lookups/esri_test.rb
+++ b/test/unit/lookups/esri_test.rb
@@ -6,6 +6,8 @@ class EsriTest < GeocoderTestCase
 
   def setup
     Geocoder.configure(lookup: :esri)
+    require 'webmock/test_unit'
+    WebMock.enable!
   end
 
   def test_query_for_geocode
@@ -19,6 +21,22 @@ class EsriTest < GeocoderTestCase
   def test_query_for_geocode_with_token_for_storage
     token = Geocoder::EsriToken.new('xxxxx', Time.now + 1.day)
     Geocoder.configure(esri: {token: token, for_storage: true})
+    query = Geocoder::Query.new("Bluffton, SC")
+    lookup = Geocoder::Lookup.get(:esri)
+    res = lookup.query_url(query)
+    assert_equal "http://geocode.arcgis.com/arcgis/rest/services/World/GeocodeServer/find?f=pjson&forStorage=true&outFields=%2A&text=Bluffton%2C+SC&token=xxxxx",
+      res
+  end
+
+  def test_query_for_geocode_with_client_credentials_for_storage
+    token_data = {
+      access_token: "xxxxx",
+      expires_in: 7200
+    }
+    stub_request(:post, "https://www.arcgis.com/sharing/rest/oauth2/token").
+      to_return(:body => token_data.to_json)
+
+    Geocoder.configure(esri: {api_key: ['id','secret'], for_storage: true})
     query = Geocoder::Query.new("Bluffton, SC")
     lookup = Geocoder::Lookup.get(:esri)
     res = lookup.query_url(query)
@@ -104,5 +122,7 @@ class EsriTest < GeocoderTestCase
 
   def teardown
     Geocoder.configure(esri: {token: nil, for_storage: nil})
+    WebMock.reset!
+    WebMock.disable!
   end
 end
--- a/test/unit/lookups/esri_test.rb
+++ b/test/unit/lookups/esri_test.rb
@@ -2,12 +2,26 @@
 require 'test_helper'
 require 'geocoder/esri_token'
 
+class Geocoder::EsriToken
+  class << self
+
+    alias_method :old_generate_token, :generate_token; 
+
+    # Stub the generate_token method when called with a specific client_id and client_secret
+    def generate_token(client_id, client_secret, expires=1440)
+      if client_id == "id" and client_secret == "secret"
+        "xxxxx"
+      else
+        self.old_generate_token client_id, client_secret, expires
+      end
+    end
+  end
+end
+
 class EsriTest < GeocoderTestCase
 
   def setup
     Geocoder.configure(lookup: :esri)
-    require 'webmock/test_unit'
-    WebMock.enable!
   end
 
   def test_query_for_geocode
@@ -19,7 +33,7 @@ class EsriTest < GeocoderTestCase
   end
 
   def test_query_for_geocode_with_token_for_storage
-    token = Geocoder::EsriToken.new('xxxxx', Time.now + 1.day)
+    token = Geocoder::EsriToken.new('xxxxx', Time.now + 86400)
     Geocoder.configure(esri: {token: token, for_storage: true})
     query = Geocoder::Query.new("Bluffton, SC")
     lookup = Geocoder::Lookup.get(:esri)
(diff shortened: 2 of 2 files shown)


alexreisner:
Thanks for this. Is the Webmock stuff necessary? It doesn't look like a request is being made.

aaronpk:
The gem makes an HTTP request to arcgis.com to exchange the client_id and client_secret for an access token. That's the part of the code that was previously untested. I added webmock to stub that request to return the fake token.

lib/geocoder/lookups/esri.rb
```diff
@@ -52,7 +52,14 @@ def token
...
       elsif configuration.api_key # generate a new token if we have credentials
         token_instance = Geocoder::EsriToken.generate_token(*configuration.api_key)
-        Geocoder.configure(:esri => {:token => token_instance})
+
+        esri_config = {:token => token_instance}
+        if configuration[:for_storage]
+          esri_config[:for_storage] = configuration[:for_storage]
+        end
+        # replaces the entire :esri config section with the new object
+        Geocoder.configure(:esri => esri_config)
```
alexreisner: It looks like you're overwriting the whole ESRI config here. What if someone sets something other than `:for_storage` or `:token`?

aaronpk: The existing code was already overwriting the whole config, which is what was causing the problem. From what I can tell, there aren't any other options that can be provided in the config other than `token` and `for_storage`.

alexreisner:
Is it possible to make this test more targeted? It's about query URL generation, so token generation shouldn't be involved. Maybe just set a token manually?

lib/geocoder/lookups/esri.rb
alexreisner: Not now, but more options could become relevant in the future. It's just not a good idea. I think there's probably a way to merge `:for_storage` into the existing configuration.

aaronpk: Okay, I can try to find a way to access the original `:esri` config options to merge them.

aaronpk:
The reason this bug exists is because the code inside the `elsif configuration.api_key` clause was never being called from any test. In order to run that code, the geocoder needs to be configured with an `api_key` and **not** a token, which means the geocoder itself is responsible for obtaining the value that gets set in the token parameter in the URL. My solution to that was to mock the HTTP call that the geocoder makes that turns the api key into the token.

lib/geocoder/lookups/esri.rb
```diff
@@ -52,7 +52,14 @@ def token
         configuration[:token].to_s
       elsif configuration.api_key # generate a new token if we have credentials
         token_instance = Geocoder::EsriToken.generate_token(*configuration.api_key)
-        Geocoder.configure(:esri => {:token => token_instance})
```
TrangPham: You can just change this to `Geocoder.configure(:esri => configuration.merge({:token => token_instance}))`

Tested via commandline
``` ruby
2.3.0 :007 > Geocoder.configure(:ersi => {token: '123'})
 => {:timeout=>3, :lookup=>:google, :ip_lookup=>:freegeoip, :language=>:en, :http_headers=>{}, :use_https=>false, :http_proxy=>nil, :https_proxy=>nil, :api_key=>nil, :cache=>nil, :cache_prefix=>"geocoder:", :basic_auth=>{}, :logger=>:kernel, :kernel_logger_level=>2, :always_raise=>[], :units=>:mi, :distances=>:linear, :ersi=>{:token=>"123"}}
2.3.0 :008 > Geocoder.config_for_lookup(:ersi)
 => {:timeout=>3, :lookup=>:google, :ip_lookup=>:freegeoip, :language=>:en, :http_headers=>{}, :use_https=>false, :http_proxy=>nil, :https_proxy=>nil, :api_key=>nil, :cache=>nil, :cache_prefix=>"geocoder:", :basic_auth=>{}, :logger=>:kernel, :kernel_logger_level=>2, :always_raise=>[], :units=>:mi, :distances=>:linear, :token=>"123"}
2.3.0 :009 > Geocoder.configure(:ersi => Geocoder.config_for_lookup(:ersi).merge({forStorage: true}))
 => {:timeout=>3, :lookup=>:google, :ip_lookup=>:freegeoip, :language=>:en, :http_headers=>{}, :use_https=>false, :http_proxy=>nil, :https_proxy=>nil, :api_key=>nil, :cache=>nil, :cache_prefix=>"geocoder:", :basic_auth=>{}, :logger=>:kernel, :kernel_logger_level=>2, :always_raise=>[], :units=>:mi, :distances=>:linear, :ersi=>{:timeout=>3, :lookup=>:google, :ip_lookup=>:freegeoip, :language=>:en, :http_headers=>{}, :use_https=>false, :http_proxy=>nil, :https_proxy=>nil, :api_key=>nil, :cache=>nil, :cache_prefix=>"geocoder:", :basic_auth=>{}, :logger=>:kernel, :kernel_logger_level=>2, :always_raise=>[], :units=>:mi, :distances=>:linear, :token=>"123", :forStorage=>true}}
2.3.0 :010 > Geocoder.config_for_lookup(:ersi)
 => {:timeout=>3, :lookup=>:google, :ip_lookup=>:freegeoip, :language=>:en, :http_headers=>{}, :use_https=>false, :http_proxy=>nil, :https_proxy=>nil, :api_key=>nil, :cache=>nil, :cache_prefix=>"geocoder:", :basic_auth=>{}, :logger=>:kernel, :kernel_logger_level=>2, :always_raise=>[], :units=>:mi, :distances=>:linear, :token=>"123", :forStorage=>true}
2.3.0 :011 > Geocoder.config_for_lookup(:google)
 => {:timeout=>3, :lookup=>:google, :ip_lookup=>:freegeoip, :language=>:en, :http_headers=>{}, :use_https=>false, :http_proxy=>nil, :https_proxy=>nil, :api_key=>nil, :cache=>nil, :cache_prefix=>"geocoder:", :basic_auth=>{}, :logger=>:kernel, :kernel_logger_level=>2, :always_raise=>[], :units=>:mi, :distances=>:linear}
```

test/unit/lookups/esri_test.rb
```diff
@@ -26,6 +28,23 @@ def test_query_for_geocode_with_token_for_storage
       res
   end
 
+  def test_query_for_geocode_with_client_credentials_for_storage
+    token_data = {
+      access_token: "xxxxx",
+      expires_in: 7200
+    }
+    stub_request(:post, "https://www.arcgis.com/sharing/rest/oauth2/token").to_return(:body => token_data.to_json)
```
TrangPham: Instead of requiring a dependency for this test, stubbing one layer higher could also work. Please try stubbing this method: `Geocoder::EsriToken.generate_token(*configuration.api_key)`

TrangPham:
Hi @aaronpk 

Thanks so much for your PR! That was a hard bug to see in the code. I've added some comments inline as well.

---
For reference, this was hard to since it looked like token was being set, returned and then for storage was being set here: https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookups/esri.rb#L45-47
What's going on here is there is the configuration of a specific lookup and then parameters for each search. When a new token is needed the global configuration was overwritten causing the for storage value to be unset. 

Here is an example of how geocoder is sending a request with params:
 ```
Geocoder.configure(:google => {:apple = true, :banana = false, :orange = true})
Geocoder.search("123 Fake Street", :params => {:banana = true})
```
Normally this would have resulted in calling 
`http://someurl.com/search?apple=true&banana=true&orange=true` 
In the case of this bug, since the global configuration was overwritten it would have resulted in calling
`http://someurl.com/search?token='sometoken'&banana=true`

lib/geocoder/lookups/esri.rb
aaronpk: Thanks! That should work fine. However, it looks like the config state isn't being reset between tests so all the tests that don't use a token are failing now. I'll keep working on it.

TrangPham: This is probably something we do want in every test - to have the state being clean before we run it.

alexreisner:
A few things:

1) For merging the token into ESRI-specific config, please use something like this:

    old_esri_config = Geocoder.config[:esri]
    Geocoder.configure(:esri => .merge({:token => token_instance}))

That way we're not overriding any global config in the ESRI-specific hash.

2) Can't the token request be preventing by simply configuring a token at the beginning of the test?

3) Here's a suggested refactoring of the `token` method that will allow for easier stubbing (if stubbing is ever necessary), more targeted tests, and improved readability:

    def token
      if (!configuration[:token] or !configuration[:token].active?) and configuration.api_key
        fetch_and_save_token!
      end
      configuration[:token].to_s
    end

    def fetch_and_save_token!
      token_instance = Geocoder::EsriToken.generate_token(*configuration.api_key)
      Geocoder.configure(:esri => {:token => token_instance})
    end

aaronpk:
> Can't the token request be preventing by simply configuring a token at the beginning of the test?

We need to test the token generation, since the current bug is caused by not having a test that covers that part of the code.

test/unit/lookups/esri_test.rb
aaronpk: Is it really a problem to have this stub request? I added this because the test suite was already using `stub_request` in a different file, so I didn't introduce a new dependency.

lib/geocoder/lookups/esri.rb
aaronpk: I ended up using @alexreisner's suggestion of `Geocoder.config[:esri].merge({:token => token_instance})` and the tests are passing now!

alexreisner:
Agree that we should test token generation, but that should be done in a method with a name like `test_token_generation`, not `test_query_for_geocode_with_client_credentials_for_storage`. The more targeted (specific) tests are, the less often they'll break when unrelated parts of the code change (extra important on an open source project).

test/unit/lookups/esri_test.rb
aaronpk: @TrangPham what mechanism can I use to stub that method? It looks like `Test::Unit` doesn't have a `.stub` method like RSpec or MiniTest do.

aaronpk:
@alexreisner @TrangPham Okay, I separated those two tests out. Hope this works for you!

alexreisner:
Sorry to do this, but this discussion is getting hard to follow (comments in too many places). I just committed 069917c039ef6f9d78e84899f31637908c83fb21 which fixes the issue named in the title of this PR. (Note that I also changed the test assertion to check specifically for the two parameters being configured. The rest of the string is irrelevant. This will prevent the test from breaking in the event something else changes about how URLs are generated.)

The token generation testing is a separate issue. @aaronpk could you open a new PR with the code you've added for that, and one test that very specifically tests that functionality?

alexreisner:
Note that I also added 1309ea63e9f9de56679f7c07027b487f92259a4d which should make testing easier.

aaronpk:
@alexreisner The test in 069917c039ef6f9d78e84899f31637908c83fb21 doesn't actually test that the new config is merged properly. The line that you changed to merge the config isn't ever hit by this test. This is demonstrated by the fact that your second commit un-does that "config merge" change but the tests still pass: https://github.com/alexreisner/geocoder/commit/1309ea63e9f9de56679f7c07027b487f92259a4d#diff-69b726dcd427393154984fe539912caeR61

I'll submit a new PR to fix this up.

alexreisner:
Oops! You're totally right about that--I was working too quickly/carelessly. Yes, please submit one PR to test token generation and another to fix my sloppiness. Thanks!

aaronpk:
Thanks, here are the two PRs: #1053 #1054

PR #1051, opened 2016-05-21
Add Mapzen reverse geocoding

### LINKED ISSUE #1039: Mapzen reverse geocoding param error
With version 1.3.4 of **geocoder**, when trying to perform a reverse geocode using **:mapzen** as lookup engine, it returns a 400 error.

Inspecting the created URL, the coordinates are included in a `text` parameter:

https://search.mapzen.com/v1/reverse?api_key=search-xxx&size=1&text=[39.57,+-0.48]

This request returns the following error:

> _parameters point.lat and point.lon must both be specified_

According to [Mapzen docs](https://mapzen.com/documentation/search/reverse/), it should be set as separated `point.lat` and `point.lon` parameters:

https://search.mapzen.com/v1/reverse?api_key=search-xxx&size=1&point.lat=39.57&point.lon=-0.48

shinyaK14:
I  implemented Mapzen reverse geocoding.

To resolve #1039 Mapzen reverse geocoding param error.

Thanks.

### DIFF
--- a/lib/geocoder/lookups/pelias.rb
+++ b/lib/geocoder/lookups/pelias.rb
@@ -23,11 +23,19 @@ module Geocoder::Lookup
     private
 
     def query_url_params(query)
-      {
+      params = {
         api_key: configuration.api_key,
-        text: query.text,
         size: 1
-      }.merge(super(query))
+      }.merge(super)
+
+      if query.reverse_geocode?
+        lat,lon = query.coordinates
+        params[:'point.lat'] = lat
+        params[:'point.lon'] = lon
+      else
+        params[:text] = query.text
+      end
+      params
     end
 
     def results(query)
--- a/test/unit/lookups/pelias_test.rb
+++ b/test/unit/lookups/pelias_test.rb
@@ -21,4 +21,10 @@ class PeliasTest < GeocoderTestCase
     query = Geocoder::Query.new('Madison Square Garden, New York, NY')
     assert_match 'size=1', query.url
   end
+
+  def test_query_for_reverse_geocode
+    lookup = Geocoder::Lookup::Mapzen.new
+    url = lookup.query_url(Geocoder::Query.new([45.423733, -75.676333]))
+    assert_match(/point.lat=45.423733&point.lon=-75.676333&size=1/, url)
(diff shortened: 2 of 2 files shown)


alexreisner:
Thanks! This code looks good but I think it should be added on the `:pelias` lookup rather than `:mapzen` (?).

shinyaK14:
@alexreisner Yes, I agree. I moved Mapzen reverse geocoding to Pelias.
The code is much better. Thanks!!

alexreisner:
OK, great. This looks good. My only other question is: why are we limiting to HTTPS? It seems like someone might want to configure an endpoint that doesn't support SSL.

shinyaK14:
@alexreisner The reason is that I couldn't find Mapzen supports http protocol in their document. 
But yeah. We should NOT limit it. In case someone want to configure as http.

alexreisner:
OK, thanks. One last thing: I think the test should be on the Pelias lookup as well, since Mapzen inherits from Pelias and that's where the reverse geocoding was implemented.

If that makes sense, please also squash all of the above into a single commit and I'll merge. Thanks!

alexreisner:
OK, thanks. One last thing: I think the test should be on the Pelias lookup as well, since Mapzen inherits from Pelias and that's where the reverse geocoding was implemented.

If that makes sense, please also squash all of the above into a single commit and I'll merge. Thanks!

alexreisner:
OK, thanks. One last thing: I think the test should be on the Pelias lookup as well, since Mapzen inherits from Pelias and that's where the reverse geocoding was implemented.

If that makes sense, please also squash all of the above into a single commit and I'll merge. Thanks!

alexreisner:
OK, thanks. One last thing: I think the test should be on the Pelias lookup as well, since Mapzen inherits from Pelias and that's where the reverse geocoding was implemented.

If that makes sense, please also squash all of the above into a single commit and I'll merge. Thanks!

shinyaK14:
@alexreisner Thanks for your advice. I really appreciate it!!
I'm a big fan of Geocoder :)

alexreisner:
Thank **you** for the contribution!

PR #1053, opened 2016-05-25
adds test for esri token generation

aaronpk:
there was previously no test of configuring the Esri geocoder with a client ID and secret. this adds a test for that, with a stub `generate_token` method to prevent the test from making an HTTP request to Esri.

Note that the test fails right now because of the bug in #1047 and re-introduced by https://github.com/alexreisner/geocoder/commit/1309ea63e9f9de56679f7c07027b487f92259a4d#diff-69b726dcd427393154984fe539912caeR61

### DIFF
--- a/test/unit/lookups/esri_test.rb
+++ b/test/unit/lookups/esri_test.rb
@@ -2,6 +2,22 @@
 require 'test_helper'
 require 'geocoder/esri_token'
 
+class Geocoder::EsriToken
+  class << self
+
+    alias_method :old_generate_token, :generate_token; 
+
+    # Stub the generate_token method when called with a specific client_id and client_secret
+    def generate_token(client_id, client_secret, expires=1440)
+      if client_id == "id" and client_secret == "secret"
+        "xxxxx"
+      else
+        self.old_generate_token client_id, client_secret, expires
+      end
+    end
+  end
+end
+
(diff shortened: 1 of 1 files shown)


alexreisner:
At its core, this PR seems to really be about merging configuration. I just created #1055 which accomplishes the same thing and packages the functionality for reuse. For clarity, I've separated the functionality of the old `token` method even further. Let me know what you think.

Also, given #1055, if you want to add a test for the esri lookup specifically in the case where an api_key (and no token) is configured, you can do so by simply stubbing the `create_token` method on the ESRI lookup instance in the test (using `instance_eval` or something similar). This way we can avoid stubbing (and then not unstubbing) important methods in the code.

Does that all make sense?

aaronpk:
@alexreisner that looks good. I do still want to add this test, since it's already broken a couple times without the test in place, and that part of the code was showing up red in a coverage report. I can re-create this after #1055 is merged.

alexreisner:
Sounds good! I just merged #1055. And I agree that that code should be tested. Might make sense to replace `test_query_for_geocode_with_token_and_for_storage` since that's no longer as relevant as it was.

aaronpk:
Great! I just added a new test in #1056, we can refactor the tests later but I don't mind having a few "extra" tests.

PR #1055, opened 2016-05-25
Add Method for Merging Lookup Configuration

alexreisner:
This functionality is currently implemented by the ESRI lookup but is potentially useful to other lookups as well.


TrangPham:
Nice :) That makes it easier to read

PR #1056, opened 2016-05-25
adds test for esri token generation from client credentials

aaronpk:
This stubs the `generate_token` method only within this test. This brings the test coverage of `Geocoder::Lookup::Esri` closer to 100%.

### DIFF
--- a/test/unit/lookups/esri_test.rb
+++ b/test/unit/lookups/esri_test.rb
@@ -29,18 +29,16 @@ class EsriTest < GeocoderTestCase
   def test_query_for_geocode_with_client_credentials_and_for_storage
     Geocoder.configure(esri: {api_key: ['id','secret'], for_storage: true})
 
-    Geocoder::EsriToken.instance_eval do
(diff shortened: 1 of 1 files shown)


alexreisner:
> ...stubbing the `create_token` method on the ESRI lookup instance in the test (using `instance_eval` or something similar). This way we can avoid stubbing (and then not unstubbing) important methods in the code.

What I mean is something like this:

    lookup.instance_eval do
      # redefine `create_token` to return a manually-created token
    end

We don't want to touch any classes, only the particular instance we're testing.

aaronpk:
Got it, fixed.

alexreisner:
Looks good! Just two things and this is ready to merge:

1. Please rename the test to explain more precisely what is being tested (eg: `test_token_generation_doesnt_overwrite_existing_config`).
2. Squash into a single commit.

Thanks for your patience!

aaronpk:
Thanks! It looks like you can squash the commits when you merge the PR. https://help.github.com/articles/about-pull-request-merge-squashing/

PR #1058, opened 2016-05-26
add configuration parameter to send a POST request

aaronpk:
since the Esri batch geocoder can generate requests that are potentially very large, it's better to set the request method to POST for those. this adds a configuration option so GET/POST can be used on any provider.

### DIFF
--- a/lib/geocoder/configuration.rb
+++ b/lib/geocoder/configuration.rb
@@ -56,6 +56,7 @@ module Geocoder
       :language,
       :http_headers,
       :use_https,
+      :use_post,
       :http_proxy,
       :https_proxy,
       :api_key,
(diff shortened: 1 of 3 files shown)


alexreisner:
Hmm... I don't think this is necessary, or particularly useful. It's also kind of misleading because I don't think any of the current APIs support POST requests, so setting this to `true` would break things. Is there a use case I'm not thinking of? (other than with ESRI batch)

aaronpk:
I only confirmed that it works with Esri. The problem is GET requests fail pretty quickly with Esri when doing batch requests, since packing that much data into the query string surpasses the limit. If you'd prefer, I can add this as an option specifically on the Esri provider, but it is definitely necessary there.

alexreisner:
I think it could be an option when using ESRI batch, but if there's no benefit to using GET then maybe on that lookup it shouldn't even be an option (it always uses POST). I've envisioned ESRI batch as a separate lookup from single-address ESRI. (Sorry, I haven't had a chance to review your PR for that yet.)

aaronpk:
There isn't any benefit to using a GET request with Esri since their API works with GET and POST, the only difference being GET has a limit on the length of the query string. I'll change it to always send a POST request for Esri.

alexreisner:
Only for the batch request, right? I think we should leave the single requests as GET.

aaronpk:
Frankly, if a provider *does* support POST requests, then we should *always* be sending a POST request to them if the request includes a token. Passing tokens in query strings is generally not a good idea, since it's far more likely for the token to be logged at various points throughout the process. I would rather just make all the Esri requests send via POST for this reason.

alexreisner:
OK. One concern I have is that it may complicate testing significantly. I'm skeptical, but happy to discuss a PR if you want to submit one.

PR #1061, opened 2016-05-31
try Rack::Request if ActionDispatch::Request is not defined, fixes #1060

### LINKED ISSUE #1060: request.location is nil in 1.3.5
Accessing the `request.location` object from a Sinatra GET handler (which returns a Sinatra::Request, which inherits from Rack::Request), throws an error.

This works fine in `1.3.4`

```
11:22:55 web.1  | 2016-05-30 11:22:55 - NoMethodError - undefined method `location' for #<Sinatra::Request:0x007ff32e027f78>:
```

alepore:
Can we test correct module inclusion in some way?

### DIFF
--- a/lib/geocoder/request.rb
+++ b/lib/geocoder/request.rb
@@ -80,4 +80,6 @@ end
 
 if defined?(ActionDispatch::Request)
   ActionDispatch::Request.include Geocoder::Request
+elsif defined?(Rack::Request)
+  Rack::Request.include Geocoder::Request
 end


alexreisner:
Thanks!

And yes, it would be good to test inclusion. I think the first step would be putting the include code into a method so a test could invoke it at will, not just when `lib/geocoder/request.rb` is required. Then the tests (one for Rails, one for Rack/Sinatra) could mock ActionDispatch and Rack respectively, call the include method, and check for presence of the `location` method on the corresponding `Request` object.

PR #1075, opened 2016-06-26
Fix ambiguous 1st argument warnings in Ruby 2.3.x

stevecj:
Having a first argument to a method call without parentheses in Ruby 2.3.x that is a "/" delimited regular expression literal results in…

warning: ambiguous first argument; put parentheses or a space even after `/' operator

This can be fixed by either using parentheses around the arguments or using the alternative syntax for literal regular expressions. It seems cleanest to me to use the alternative syntax rather than adding the parentheses when similar nearby code lines all do not use the parentheses.


alexreisner:
Thanks!

PR #1076, opened 2016-06-26
Use sqlite ruby math extensions if present

stevecj:
I think this is basically ready to incorporate, but…
* Is the travis setup right?
* Are the Gemfile changes right?
Maybe the CI run for this PR will tell us those things.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,3 +46,5 @@ matrix:
       gemfile: gemfiles/Gemfile.ruby1.9.3
     - rvm: 2.3.0
       gemfile: gemfiles/Gemfile.ruby1.9.3
+    - env: DB=sqlite USE_SQLITE_EXT=1
+      rvm: jruby-19mode
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ env:
   matrix:
     - DB=
     - DB=sqlite
+    - DB=sqlite USE_SQLITE_EXT=1
     - DB=postgres
     - DB=mysql
 rvm:
--- a/Gemfile
+++ b/Gemfile
@@ -24,6 +24,7 @@ end
 
 group :test do
   gem 'sqlite3', :platform => [:ruby, :mswin, :mingw]
+  gem 'sqlite_ext', '~> 1.4.0', :platform => [:ruby, :mswin, :mingw]
   gem 'webmock'
 
   platforms :ruby do
--- a/Gemfile
+++ b/Gemfile
@@ -23,8 +23,11 @@ group :development, :test do
 end
 
 group :test do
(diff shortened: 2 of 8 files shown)


alexreisner:
> ...failing CI, presumably because the sqlite_ext gem requires Ruby 2.x. Can anyone help me figure out what to put in the .travis file to have the "DB=sqlite USE_SQLITE_EXT=1" matrix entry only be used with applicable Ruby versions?

@TrangPham could you help with this? You're more familiar with the Travis setup than I am.

stevecj:
I've decided the better way to handle this will be to simply make sqlite_ext compatible with older Ruby versions. I can use the alias method chain pattern to make that work.

stevecj:
I have broadened the sqlite_ext gem for version 1.5.0 to support older Ruby versions and updated the "Gemfile"s in geocoder to require that version or newer. Let's see if CI likes it now. :)

stevecj:
It looks like the tests are passing now. There are still a couple of things to confirm before merging though:
1. I added `- DB=sqlite USE_SQLITE_EXT=1` to the "matrix" array in .travis.yml, but does that do what I expect? What I expect is for that to result in the `DB` environment variable containing "sqlite" and the `USE_SQLITE_EXT` environment variable containing "1".
2. The `DB=sqlite USE_SQLITE_EXT=1` configuration is now being run for platforms in which it will have no effect, which does no specific harm, but wastes time and resources in CI. Is there any way to omit those runs for non-applicable platforms?

stevecj:
Ah. Waiting a while longer, I'm seeing failures now in the jruby runs. Testing is trying to use sqlite_ext whenever `USE_SQLITE_EXT=1` even if sqlite_ext is not loaded because it's not appropriate for the platform. I'll fix that.

That actually answers question 1 in my comment above as well. jruby runs without `USE_SQLITE_EXT=1` are passing and jruby runs with that are failing, so the environment variable must be getting set.

stevecj:
Several failed tries & force-pushes later, I think that I have the CI testing in excellent condition now. :)

alexreisner:
This is looking pretty good! It's certainly minimal in terms of added complexity. In fact, I think the only changes I'd recommend are in `test_helper.rb`:

1. Rename `using_raw_sqlite?` to `using_unextended_sqlite?` ("raw" is less specific).
2. Is there a way that method can use the existing extension-detecting code in the ActiveRecord module?
3. Remove "_without_raw_sqlite" from test names (the `omit` line makes this clear enough).

Thanks for your patience with Travis, by the way.

stevecj:
1 & 3: Will do.
2: I'll look into that & comment about what I learn.

alexreisner:
Hey, sorry for the delay. I didn't get a notification when you added the commit. Merging now. Thanks!

PR #1078, opened 2016-06-27
more status codes from Google Maps API

adlersantos:
This adds more status codes as shown in the Google Maps API Documentation (see status codes [here](https://developers.google.com/maps/documentation/geocoding/intro#StatusCodes)):

- `ZERO_RESULTS` indicates that the geocode was successful but returned no results. This may occur if the geocoder was passed a non-existent address.
- `UNKNOWN_ERROR` indicates that the request could not be processed due to a server error. The request may succeed if you try again.

Hope you can check! :)

### DIFF
--- a/lib/geocoder/exceptions.rb
+++ b/lib/geocoder/exceptions.rb
@@ -25,6 +25,12 @@ module Geocoder
   class InvalidRequest < Error
   end
 
+  class ZeroResults < Error
+  end
+
+  class UnknownError < Error
+  end
+
   class InvalidApiKey < Error
   end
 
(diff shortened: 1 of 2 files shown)


alexreisner:
Thanks for this, but I don't think these are errors we need to support: 

* If a developer needs to handle the case where no results are returned they can just check `results.size == 0` rather than raising an exception.
* The unknown error might be useful, but in the context of a Ruby app it should be called something like `TemporaryServerError`. However, Google is not clear on whether it's temporary ("*may* succeed if you try again") so I'm a little hesitant to label it as such.

Let me know your thoughts. Do you have specific needs for these in your app?

adlersantos:
Thanks for the comment. My use case is that when a host sets up a property (vacation rental), he/she first supplies a street address, postal code, and country code. These will be used for geocoding. 

If something about the request went wrong, it would be appropriate to inform the host that there's a problem that's not her fault. But if the request is successful but no results were found, it's better to tell the host that she needs to check if the street address, postal code, and country code she's given are correct.

For the `ZERO_RESULTS` error, I agree to simply check if there are no results via its size.

I think `UnknownError` is still something vaguely defined by Google, and might be broken down into more specific errors in the future. Closing this one then!

Issue #1070, opened 2016-06-21
Add suppport for Location IQ


tinderfields:
http://locationiq.org/ is a new service which offers a generous 30k limit per day on their free plan.

Seems a good provider to support.

TrangPham:
Hi, we welcome you to add a pull-request for this :)

Please look at this PR as an example: https://github.com/alexreisner/geocoder/pull/997

gopi-ar:
Thanks for adding us to the list, @tinderfields ! We don't have any ruby devs on our team to add this PR, though. @TrangPham ours is a much faster version of the OSM Nominatim API, updated daily. The API specs are exactly the same, so perhaps you could just clone the providers?

lime:
@gopi-ar That's good info for anyone who wants to give it a try. :)

It might even make sense to subclass the `Nominatim` provider, the same way as e.g. `Dstk` subclasses `Google`: [lookups/dstk.rb](https://github.com/alexreisner/geocoder/blob/v1.3.6/lib/geocoder/lookups/dstk.rb)

If you want to give it a try, Geocoder is in my opinion one of those libraries that should be easy to approach even for a first-time rubyist. ;) And the maintainers are really helpful in guiding you through any quirks you encounter in your PRs.

alexreisner:
See also, as a guide: c278f94cb096215283bb35e52901159cb1e311ac

gopi-ar:
@lime @alexreisner our intern @aleemuddin13 submitted a PR integrating with LocationIQ. He had a lot of difficulty writing tests (because OSM doesn't require an API key and we do), would appreciate any help from you folks to help finish the PR.

jstenhouse:
Any chance we could get a release cut to include the new LocationIQ lookup? I've been using the master version for a couple weeks now and it's been working great with LocationIQ for my site.

alexreisner:
Resolved by #1088

PR #1088, opened 2016-08-25
Added locationiq provider

aleemuddin13:
Add support for LocationIQ #1070 

I don't know how to write tests. I would appreciate if any one help me out.


TrangPham:
Hi @aleemuddin13 

Since your API is similar, you can use this to base your test file off of: https://github.com/alexreisner/geocoder/blob/master/test/unit/lookups/nominatim_test.rb

There are also some tests which you will have to modify. You can look at the travis failures to see which tests need to be fixed. This is because some of the tests loop through all the different lookups. 

To run the tests locally with default environment variables you can run `rake test` in your terminal. Travis would run the entire test suite for you with the different environment variables, so you'll want to make sure that all the travis tests are also passing after you have fixed the local tests. 

Please reach out if you have more specific questions or are blocked.
Thanks!

aleemuddin13:
Thanks to @glsignal, tests are done.

aleemuddin13:
@alexreisner, naming conventions updated.

alexreisner:
Looks great, thanks!

jgdreyes:
@alexreisner When do you think you will push out a new release with this change?

PR #1092, opened 2016-09-05
Fix result of ipinfo.io for loopback address (127.0.0.1)

teohm:
Patch https://github.com/alexreisner/geocoder/pull/997

When query loopback address using ipinfo.io,
```
Geocoder.search("127.0.0.1")
```

This error is raised because an invalid result is returned:

```
undefined method `split' for nil:NilClass
geocoder (1.3.7) lib/geocoder/results/ipinfo_io.rb, line 11
```

Changes:
- Returns a valid result when query loopback address in ipinfo.io
- Fix incorrect method definition for extra result attributes i.e. `result.ip`, `result.region`, `result.postal`

### DIFF
--- a/lib/geocoder/lookups/ipinfo_io.rb
+++ b/lib/geocoder/lookups/ipinfo_io.rb
@@ -42,7 +42,14 @@ module Geocoder::Lookup
     end
 
     def reserved_result(ip)
-      {"message" => "Input string is not a valid IP address", "code" => 401}
+      {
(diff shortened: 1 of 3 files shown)


teohm:
@alexreisner Can help to take a look? 👀  🙇

alexreisner:
Thanks for this!

teohm:
👍 💛 😋

PR #1093, opened 2016-09-05
support canada for geocodio

bolandrm:


alexreisner:
Thanks for this! Please just fix the failing test and I'll merge.

bolandrm:
@alexreisner  I fixed this.  It's not really an awesome solution though because geocodio doesn't return the postal code for canadian results.

alexreisner:
I think that's the best we can do, though. Thanks!

PR #1096, opened 2016-09-16
improved performance of various calculations

austinthecoder:
Instead of converting the Earth's radius to different units on-the-fly, it's calculated once on file load. Here's a before/after benchmark of `distance_between`:

``` ruby
Benchmark.measure do 
  1_000_000.times do 
    Geocoder::Calculations.distance_between [47.858205,2.294359], [40.748433,-73.985655]
  end
end

              user     system      total         real
before:   7.840000   0.010000   7.850000 (  7.853178)
after:    5.800000   0.010000   5.810000 (  5.821909)
```

The changes are backwards compatible. These aren't used internally and can be deprecated if you like:

```
EARTH_RADIUS
#to_kilometers
#to_miles
#to_nautical_miles
#mi_in_km
#km_in_mi
#km_in_nm
#nm_in_km
```

### DIFF
--- a/lib/geocoder/calculations.rb
+++ b/lib/geocoder/calculations.rb
@@ -11,12 +11,6 @@ module Geocoder
     COMPASS_POINTS = %w[N NE E SE S SW W NW]
 
     ##
-    # Radius of the Earth, in kilometers.
-    # Value taken from: http://en.wikipedia.org/wiki/Earth_radius
-    #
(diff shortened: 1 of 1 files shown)


alexreisner:
Thanks for this! I like it as much for the cleaner code as for the performance improvement.

alexreisner:
I've also deprecated the methods you mentioned. Thanks again.

PR #1103, opened 2016-10-04
Update CHANGELOG.md

jaredbeck:
Reassure people that upgrading to 1.4.0 is safe by explicitly stating that there are no breaking changes.

[ci skip]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,13 @@ Major changes to Geocoder for each release. Please see the Git log for complete 
 
 1.4.0 (2016 Sep  8)
 -------------------
+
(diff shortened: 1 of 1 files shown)


alexreisner:
The problem with this is that the first item in the list (change in definition of `geocoded`) could cause different results in some apps. It's a "fix," but we waited until 1.4.0 to release it because it could (in some small percentage of apps) break things.

jaredbeck:
I was wondering about that one. I skimmed over the issue and it seemed like a "fix" to me :)

Maybe we should call it a "breaking change" then, just to draw attention to it. I'll open another PR.

Issue #1080, opened 2016-07-01
Adhere to semantic versioning


PragTob:
Hi there,

thanks for the geocoding gem - I've used it on a number of occasions and quite liked it :)

Looking to update it in a bigger application today, I glanced at the [Changelog](https://github.com/alexreisner/geocoder/blob/master/CHANGELOG.md) today and was a bit surprised by the breaking changes happening in what are patch level updates in semver. 

* 1.2.4 dropped a whole geocoding service
* 1.2.7 dropped support for all of ruby 1.9.2

I'd propose that going forward, versions would/could be semver-ish. I mean optimally those would be in major version bumps (imo). Personally I could also live with them happening in minor version bumps, but imo patch level updates should be reserved for fixes - i.e. updating from 1.2.x to 1.2.y should be doable without any fear.

What do you think?

alexreisner:
@PragTob I totally agree. I was a little careless in removing Ruby 1.9.2 support. That probably should have been done in 1.3.0. Removing support for Coudmade in 1.2.4 (and later for Yahoo) is not quite the same thing, though, because the service was shut down, rendering the code useless/misleading (anyone using it had to change their lookup anyway, regardless of gem version). That kind of thing will continue to happen and I don't think it warrants a major/minor version increase. But I agree that we should be more careful with other aspects of the gem.

PR #1104, opened 2016-10-04
Warn people about minor breaking change

jaredbeck:
Per conversation in https://github.com/alexreisner/geocoder/pull/1103

[ci skip]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,13 @@ Major changes to Geocoder for each release. Please see the Git log for complete 
 
 1.4.0 (2016 Sep  8)
 -------------------
+
+Minor breaking change
+
 * Only consider an object geocoded if both lat and lon are present (previously was sufficient to have only one of the two) (thanks github.com/mltsy).
+
+Fixed
+
 * Add support in :geocodio lookup for Canadian addresses (thanks github.com/bolandrm).
 * Add support for SQLite extensions, if present (thanks github.com/stevecj).
 * Always include Geocoder in Rack::Request, if defined (thanks github.com/wjordan).


alexreisner:
Thanks for this, but I just don't think it's necessary. This isn't really a breaking change, it just might affect the results of certain queries in a really tiny percentage of apps (ie, irrelevant to the vast majority of devs), and I think the minor version number increase is enough to alert people to that. It might be worth coming up with a way of indicating changes that could affect results, but I don't want to change the formatting of the CHANGELOG to do so (requires going back and updating for all previous releases), and honestly there are so many ways that changes affect apps that this is a bit of a rabbit hole. Eg: if a geocoding API is shut down and we remove it from Geocoder, is that a breaking change? More discussion on #1080.

jaredbeck:
The goal here is to help people upgrade, by clarifying how likely a change is to cause them trouble. However we accomplish this is fine by me. Even a little parenthetical note at the end of the bullet point would be better than nothing, i.e. making no distinction at all between bullet points.

> What’s the point of a change log? To make it easier for users and contributors to see precisely what notable changes have been made between each release (or version) of the project.
> http://keepachangelog.com/en/0.3.0/

The first bullet point is much more "notable" than the next three, isn't it?

Sorry for being persistent here, I'd just like to see a world where upgrading gems is easier :)

alexreisner:
I hear you, and I want the same thing. It's just that this particular item is difficult to categorize. It's somewhere in between a breaking change and a totally innocuous one. And I think the note already in the CHANGELOG "(previously was sufficient to have only one of the two)" explains it pretty well. I'm not sure it's more notable than the other three items. For the vast majority of users it will have _less_ effect than the others.

Generally speaking, I don't include all changes in the CHANGELOG. The really minor stuff is omitted to encourage people to read and understand each item. If there were 10 items for each version then I could see more of a need for some kind of tagging, but as it stands I think the existing explanation for this item is alright. If a breaking change is introduced (the only ones so far have been dropping support for old Ruby/Rails) I'll tag with something in all caps.

jaredbeck:
> It's just that this particular item is difficult to categorize.

Makes sense.

> Generally speaking, I don't include all changes in the CHANGELOG. The really minor stuff is omitted to encourage people to read and understand each item.

Yes! :+1::+1: That's great.

> If a breaking change is introduced (the only ones so far have been dropping support for old Ruby/Rails) I'll tag with something in all caps.

Awesome, thanks. I think we're on the same page :)

alexreisner:
OK, great. And thanks again for this. Primary among my concerns for Geocoder are ease of installation and lack of surprise during use, so I consider this and #1080 to be issues worth discussing.

PR #1116, opened 2016-11-09
Refactor calculations

tunamonster:
Refactored `calculations::extract_coordinates` and `calculations::coordinates_present?`

`coordinates_present?
`Checks whether an argument matches validating criteria (numeric, not NaN), rather than returning false when matching invalidating criteria.

Also replaced `.to_s == NaN` with` .to_f..nan? `

`extract_coordinates` 
Now calls `coordinates_present?` to validate the array, instead of comparing four booleans and initialising extra vars.

### DIFF
--- a/lib/geocoder/calculations.rb
+++ b/lib/geocoder/calculations.rb
@@ -38,12 +38,7 @@ module Geocoder
     # Returns true if all given arguments are valid latitude/longitude values.
     #
     def coordinates_present?(*args)
-      args.each do |a|
-        # note that Float::NAN != Float::NAN
-        # still, this could probably be improved:
-        return false if (!a.is_a?(Numeric) or a.to_s == "NaN")
-      end
-      true
+      args.map { |a| (a.is_a? Numeric and !a.to_f.nan?) ? true : false }.all?
     end
 
     ##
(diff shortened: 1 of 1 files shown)


lib/geocoder/calculations.rb
```diff
@@ -38,12 +38,7 @@ module Calculations
     # Returns true if all given arguments are valid latitude/longitude values.
     #
     def coordinates_present?(*args)
-      args.each do |a|
-        # note that Float::NAN != Float::NAN
-        # still, this could probably be improved:
-        return false if (!a.is_a?(Numeric) or a.to_s == "NaN")
-      end
-      true
+      args.map { |a| (a.is_a? Numeric and !a.to_f.nan?) ? true : false }.all?
```
lime: Just a drive-by suggestion for simplifying this further:
  - The expression is already a boolean, the ternary `? true : false` is not needed
  - [`Enumerable#all?`](https://ruby-doc.org/core-2.3.1/Enumerable.html#method-i-all-3F) takes a block argument:

```ruby
args.all? { |a| a.is_a? Numeric and !a.to_f.nan? }
```

alexreisner: @lime thanks for that, and I agree. @tunamonster will merge if you can make those two changes.

alexreisner:
Just a heads-up that this caused some problems and had to be reverted. See #1130 for details.

PR #1117, opened 2016-11-15
Fix mapbox results

jcmuller:
Some mapbox results come back without context. The mapbox result class was assuming that this context is always present. This patch changes that assumption.

Also, extract a couple of methods.


jcmuller:
When can we expect a new release?

Cheers!

PR #1122, opened 2016-11-25
support for opencagedata optional parameters

stemps:
I added support for a bunch of optional opencage query parameters. Thought this could be interesting for other users as well. Feedback or suggestions are very welcome.

### DIFF
--- a/lib/geocoder/lookups/opencagedata.rb
+++ b/lib/geocoder/lookups/opencagedata.rb
@@ -47,6 +47,12 @@ module Geocoder::Lookup
         :language => (query.language || configuration.language)
       }.merge(super)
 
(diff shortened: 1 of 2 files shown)


alexreisner:
It's better to declare an array of symbols than to convert strings to symbols twice. Other than that, this looks good!

stemps:
👍  good point! It's an array of symbols now.

alexreisner:
Thanks! Unfortunately I think the `%I` shortcut was added in Ruby 2.0 and we still have to support 1.9.3 a little longer... Can you write it out the old fashioned way?

stemps:
Ah... of course! I amended the commit.

PR #1125, opened 2016-11-28
Ban data gouv fr lookup

JulianNacci:
Hello,

I have integrated a new lookup base on french open data geocoding API produced by French state, called "Base Adresse Nationale".

API is reliable, used by some of top 15 french websites. They claim 200million+ geocoding requests on the service since the opening.

It has many advantages : 
- it is free
- it does not require API key
- it supports https
- it has no rate limit (from what I know)
- it is somehow better than Google at geocoding French addresses
- cool meta infos (population of cities, official UUID, etc)

Details about this API can be found here : 
[https://adresse.data.gouv.fr/api/](https://adresse.data.gouv.fr/api/) (in french)

We are using the fork I am making a PR for in production at [www.publidata.io](http://www.publidata.io).

I was planning to spend some time on the integration of another open data API edited by french state, returning more detailed information about cities, returning polygons, surface, etc : [https://docs.geo.api.gouv.fr/](https://docs.geo.api.gouv.fr/)

Feel free to comment, I am rather new at contributing on OS projects.


alexreisner:
Nice work! At first glance this looks great. I will go over it in more detail as soon as I can... Thanks for the helpful explanation.

lib/geocoder/lookups/ban_data_gouv_fr.rb
```diff
@@ -0,0 +1,130 @@
+require 'geocoder/lookups/base'
+require 'geocoder/results/ban_data_gouv_fr'
+
+module Geocoder::Lookup
+  class BanDataGouvFr < Base
+
+    def name
+      "Base Adresse Nationale Française"
```
TrangPham: This seems to create errors in loading the file. See below from Travis tests:
```
$ bundle exec rake
/home/travis/.rvm/gems/ruby-1.9.3-p551/gems/power_assert-0.4.1/lib/power_assert.rb:130: warning: (...) interpreted as grouped expression
/home/travis/.rvm/gems/ruby-1.9.3-p551/gems/power_assert-0.4.1/lib/power_assert.rb:130: warning: (...) interpreted as grouped expression
/home/travis/build/alexreisner/geocoder/test/test_helper.rb:364:in `require': /home/travis/build/alexreisner/geocoder/lib/geocoder/lookups/ban_data_gouv_fr.rb:8: invalid multibyte char (US-ASCII) (SyntaxError)
/home/travis/build/alexreisner/geocoder/lib/geocoder/lookups/ban_data_gouv_fr.rb:8: invalid multibyte char (US-ASCII)
/home/travis/build/alexreisner/geocoder/lib/geocoder/lookups/ban_data_gouv_fr.rb:8: syntax error, unexpected $end, expecting keyword_end
      "Base Adresse Nationale Française"
                                    ^
	from /home/travis/build/alexreisner/geocoder/test/test_helper.rb:364:in `<module:Lookup>'
	from /home/travis/build/alexreisner/geocoder/test/test_helper.rb:98:in `<module:Geocoder>'
	from /home/travis/build/alexreisner/geocoder/test/test_helper.rb:97:in `<top (required)>'
	from /home/travis/build/alexreisner/geocoder/test/unit/active_record_test.rb:2:in `require'
	from /home/travis/build/alexreisner/geocoder/test/unit/active_record_test.rb:2:in `<top (required)>'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/rake-11.3.0/lib/rake/rake_test_loader.rb:10:in `require'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/rake-11.3.0/lib/rake/rake_test_loader.rb:10:in `block (2 levels) in <main>'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/rake-11.3.0/lib/rake/rake_test_loader.rb:9:in `each'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/rake-11.3.0/lib/rake/rake_test_loader.rb:9:in `block in <main>'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/rake-11.3.0/lib/rake/rake_test_loader.rb:4:in `select'
	from /home/travis/.rvm/gems/ruby-1.9.3-p551/gems/rake-11.3.0/lib/rake/rake_test_loader.rb:4:in `<main>'
rake aborted!
```

JulianNacci: Ok i guess this the "ç" :/

lib/geocoder/lookups/ban_data_gouv_fr.rb
```diff
@@ -5,7 +5,7 @@ module Geocoder::Lookup
   class BanDataGouvFr < Base
 
     def name
-      "Base Adresse Nationale Française"
+      "Base Adresse Nationale Francaise"
```
JulianNacci: Ok I will do this, I also spotted this issue in other files with different characters. 

I only need to insert `# encoding: utf-8` on top of each file ? 
Is there a way to make files inherit this encoding ?
Will it be an issue in fixtures files (which also contain utf-8 chars) ?

Thanks for your assistance

JulianNacci: I have reverted commit and added `# encoding: utf-8` to the concerned files

JulianNacci:
I have made some modifications in order to make more tests pass, but I still got issue on cache hit test and no result test (I return [] if no result but it does not appear to work this way in tests).

Can you please have a look ?

alexreisner:
The test `test_no_search_results` is unnecessary. The test for no results is run on every lookup automatically, and that test runs a query for "no results". So if you remove your test and remove the word "search" from the no_search_results fixture that test should pass.

For the caching failures, you have to define a private method called `valid_response?` in your lookup that returns `true` if the response should be cached. This means the response did not timeout or result in an error. The default is to return true if the HTTP response code was in the 200-399 range, but not all APIs return proper response codes, and even when they do this is not always sufficient for determining a non-error condition. Two examples:

* Bing [looks at the HTTP status and also the headers](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookups/bing.rb#L68)
* Google [looks at the HTTP status and also the JSON data](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookups/google.rb#L44)

JulianNacci:
Thanks for insights, I have fixed no result test but still got issues on : 
- cache test :  results.first.cache_hit returns true as if cache was not empty before first request
- error handling test : error_handling_test.rb:12:in `block in test_does_not_choke_on_timeout'

JulianNacci:
I have a precision : 
I was planning to introduce geocoder gem and BAN open geocoing API implem during worldwide open government summit which will take place in Paris from https://en.ogpsummit.org/osem/conference/ogp-summit

It would be awesome to both : 
- promote ruby in the universe of democracy and civic-tech which are mainly php and python
- promote this gem as a reference on geocoding with open and free API
- get PR / visibility in those networks

Therefore it would be awesome to see it mergeable / merged /released before 6th december, do you think it would be possible / realistic ?

alexreisner:
Yep, that seems reasonable. We just need to get these tests to pass. I think the reason the cache test is failing is that `fetch_data` is called three times in the course of executing a single query, which also means you're making three times as many network requests as you need to. It should only be called once!

alexreisner:
Suggested refactoring:

    def any_result?(response)
      response['features'].any?
    end

    def results(query)
      if doc = fetch_data(query)
        if any_result?(doc)
          [doc]
        else
          []
        end
      else
        raise_error(Geocoder::Error) || warn("BAN Geocoding API error")
      end
    end

alexreisner:
And I think the timeout failure is happening because the if/else structure raises an exception if `fetch_data` returns false, which is not what should happen (`fetch_data` will raise its own errors, if any). So actually you can replace the entire `results` method with this:

    def results(query)
      if doc = fetch_data(query) and any_result?(doc)
        [doc]
      else
        []
      end
    end

(assuming you use `any_result?` from my previous comment)

alexreisner:
The final test error can be fixed by defining `state_code` in the Result object, eg: `alias_method :state_code, :state`

alexreisner:
Merged, thanks for this! Will release some time before Dec 6.

JulianNacci:
Ok, I was about to amend my last commit to fix the any_result? method which was breaking, but you were quicker :) thanks for your assistance !

PR #1134, opened 2016-12-12
Fix regression in calculations refactoring

fjl82:


lib/geocoder/calculations.rb
```diff
@@ -42,7 +42,7 @@ module Calculations
     # Returns true if all given arguments are valid latitude/longitude values.
     #
     def coordinates_present?(*args)
-      args.all? { |a| a.is_a? Numeric and !a.to_f.nan? }
+      args.all? { |a| a.is_a? Numeric or !a.to_f.nan? }
```
lime: I highly doubt that this is the fix you are looking for. This change would make `coordinates_present?` return `true` on essentially any arguments, including e.g. `Float::NAN` (which is a numeric) and any string value (which becomes a non-NaN number when converted `to_f`).

```rb
[Float::NAN, 'foo bar'].all? { |a| a.is_a? Numeric or !a.to_f.nan? }
# => true
```

TrangPham:
The change in logic to this code happened in: https://github.com/alexreisner/geocoder/commit/1c306367588f61b418a0f41f049f73cfd9f94897  We should verify that this change is included in later versions, but not versions that the error is not happening on. 

I went through it and it seems like its using demorgan correctly and that .all? Is also used correctly. 

What's a case that XOR would be true?

Also please add a test case to verify and prevent future regression.

alexreisner:
Closing this, because it's old and I agree with @lime.

PR #1140, opened 2017-01-08
Use manual batching in rake task

### LINKED ISSUE #1126: Calling rake geocode:all CLASS=YourClass throws NoMethodError
The rake task fails with the following issue, using Mongoid. Also, confirming that the library works well as part of my app, its just running it in bulk fails. (also confirming that I renamed "YourClass" when calling rake)

```
rake aborted!
NoMethodError: undefined method `find_each' for #<Mongoid::Criteria:0x007ff02855d370>
Did you mean?  find
               find_all
/Users/user/.rvm/gems/ruby-2.3.1@our-app/gems/mongoid-5.1.5/lib/mongoid/criteria.rb:502:in `method_missing'
/Users/user/.rvm/gems/ruby-2.3.1@our-app/gems/geocoder-1.4.0/lib/tasks/geocoder.rake:19:in `block (2 levels) in <top (required)>'
/Users/user/.rvm/gems/ruby-2.3.1@our-app/gems/rake-11.3.0/exe/rake:27:in `<top (required)>'
/Users/user/.rvm/gems/ruby-2.3.1@our-app/bin/ruby_executable_hooks:15:in `eval'
/Users/user/.rvm/gems/ruby-2.3.1@our-app/bin/ruby_executable_hooks:15:in `<main>'
Tasks: TOP => geocode:all
```

brendandillon:
Changes batching in the rake task to work using a loop instead of the find_each method which does not work with Mongoid. This should fix #1126.

### DIFF
--- a/lib/tasks/geocoder.rake
+++ b/lib/tasks/geocoder.rake
@@ -11,14 +11,20 @@ namespace :geocode do
     reverse = false unless reverse.to_s.downcase == 'true'
 
     if reverse
-      klass.not_reverse_geocoded.find_each(batch_size: batch) do |obj|
-        obj.reverse_geocode; obj.save
-        sleep(sleep_timer.to_f) unless sleep_timer.nil?
+      size = klass.not_reverse_geocoded.count
+      (0..size).step(batch) do |offset|
+        klass.not_reverse_geocoded.skip(offset).limit(batch).each do |obj|
+          obj.reverse_geocode; obj.save
+          sleep(sleep_timer.to_f) unless sleep_timer.nil?
+        end
       end
     else
-      klass.not_geocoded.find_each(batch_size: batch) do |obj|
-        obj.geocode; obj.save
-        sleep(sleep_timer.to_f) unless sleep_timer.nil?
+      size = klass.not_geocoded.count
+      (0..size).step(batch) do |offset|
+        klass.not_geocoded.skip(offset).limit(batch).each do |obj|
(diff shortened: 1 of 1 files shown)


alexreisner:
Thanks for this. However, I think it would be better to detect whether Mongoid or ActiveRecord is being used in the model and do whatever's appropriate in each case. Getting the same code to work for both is always going to be a pain. For example, the `skip` method won't work with ActiveRecord. And I believe [batches aren't necessary with Mongoid](https://jira.mongodb.org/browse/MONGOID-1335).

brendandillon:
Thanks for the input; I have changed it to be dependent on the ORM being used. This seems to work properly with an ActiveRecord repo as well as a Mongoid repo on my machine.

alexreisner:
Thanks! This approach is better. A few things:

1. We should check to see if the particular model includes Mongoid, rather than just checking whether Mongoid is being used anywhere in the app.
2. I think you mean to use `=` rather than `:` for the `reverse` argument of `geocode_record`.
3. I think the if/else block can be DRYed up considerably, eg:

```
scope = reverse ? klass.not_reverse_geocoded : klass.reverse_geocoded
if orm == 'mongoid'
  scope.each do |obj|
    geocode_record(obj, reverse: reverse)
  end
else
  scope.find_each(batch_size: batch) do |obj|
    geocode_record(obj, reverse: reverse)
  end
end
```

PR #1141, opened 2017-01-11
Place of interest can be referred from `place_type` for Nominatim result

zocoi:
Here's an example:
```
http://nominatim.openstreetmap.org/details.php?place_id=118227317
```

The type is restaurant which is missing in the current list but we can get it from `@data['type']`


zocoi:
hhmm, I'm not sure why build `rbx-2` was failing

alexreisner:
Thanks! And don't worry about the RBX test failures, they're unrelated.

Issue #1128, opened 2016-12-01
Geocoding with Google Place Search?


waruboy:
Hello.

First, thank you very much for the gem. We use it at storemapper.co and very happy with it.

Recently, Google announce that they will make the change to geocoding API that will make it more unsuitable for unambiguous query. For ambiguous search they reccommend using Google Place Search

> We plan to roll out an update to the Geocoding API at the end of November 2016 that will increase the difference between Geocoding and Places performance for ambiguous and unambiguous queries. This change will improve the quality of Geocoding results for unambiguous queries, but will be more likely to return ZERO_RESULTS for ambiguous or incomplete queries where the Geocoding API was unable to find a high quality result. 

https://googlegeodevelopers.blogspot.co.id/2016/11/address-geocoding-in-google-maps-apis.html

As far as I know Google Place Search (not Details) is not supported yet by the gem. Is there any plan to support it? Will be happy to help with PR.

alexreisner:
Thanks! Glad you guys are happy with the gem. Thanks also for the link to the blog post, which I hadn't seen. You're correct that Place Search is not currently supported. I'm open to a PR as long as the Lookup and Result classes follow the basic form of the existing classes (ie, the request and response formats are not significantly different, which they don't appear to be from a quick glance at the documentation).

waruboy:
Cool! let me see what I can do.

alexreisner:
Closed by #1143

PR #1143, opened 2017-01-13
Google Places Search Lookup

waruboy:
Hello. This is my attempt to integrate gogole places search lookup, as discussed on https://github.com/alexreisner/geocoder/issues/1128 . 

This query return way less data than other google query though, so many method will just return blank.
What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -436,7 +436,7 @@ The [Google Places Details API](https://developers.google.com/places/documentati
 
 #### Google Places Search (`:google_places_search`)
 The [Google Places Search API](https://developers.google.com/places/web-service/search) is the geocoding service of Google Places API.
-It only returns limited address data, but it also returns place id. You can use Google Place Details if you need more detailed information using obtained place id. See points details on Google Places Details in previous section. You should use this service instead of regular Google Geocoding API if your queries are ambiguos or incomplete addresses. For more comparison between this and regular Google Geocoding API, see https://maps-apis.googleblog.com/2016/11/address-geocoding-in-google-maps-apis.html
+It only returns limited location data, however it also returns corresponding `place_id` for a given location. You can use Google Place Details if you need more detailed information using obtained place id. See points details on Google Places Details in previous section. You should use this service instead of regular Google Geocoding API if your queries are ambiguos or incomplete addresses. For more comparison between this and regular Google Geocoding API, see https://maps-apis.googleblog.com/2016/11/address-geocoding-in-google-maps-apis.html
 
 #### Bing (`:bing`)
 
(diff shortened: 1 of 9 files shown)


waruboy:
Hmm, I have no idea why the rbx based CI test is failing. Something about the gemset which I don't touch

alexreisner:
Ah, sorry, don't worry about that failing test. Nothing to do with what you did, which looks great! I just need to find the time to go over it more carefully--at the moment the only thing I can spot that's missing is an entry in the README.

waruboy:
@alexreisner Thanks for responding! No pressure from here, I know it can be though maintaining an OSS project. Ah yes, will draft something to add in the README

alexreisner:
Looks good. One more thing: is it necessary to define the `results` method in the lookup? It looks the same as the method in the lookup it's inheriting from, but maybe I'm missing something subtle...

waruboy:
@alexreisner Ah yeah, only different is just the log message when raising error.
I followed the style of GooglePlaceDetails lookup

alexreisner:
Ah, right. In that case I think it'd be better to make the superclass's method insert the name of the lookup dynamically into the error message. There's too much critical functionality in there to copy and paste. It will have the side effect of slightly improving the Google Premier lookup's error message, too.

waruboy:
@alexreisner good idea. Should we make it another PR or just merge it here? Would be happy to do either way

alexreisner:
Here is good. Thanks!

waruboy:
@alexreisner done!

alexreisner:
Ah, sorry, I should have been more specific. Just call the `name` method on the lookup for the human-readable name that should go in the error message. That make sense?

waruboy:
@alexreisner  woops, totally forgot that we have that method. Is this what you mean?

aceunreal:
Is it worth also adding a `google_premier_places_search` lookup?

waruboy:
@aceunreal I don't know. Does it need different request? The premium data feature will be deprecated though
https://developers.google.com/places/web-service/search#premium-data

aceunreal:
@waruboy when using a Google premium plan, you have the possibility of using a signature as the authentication. eg: https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookups/google_premier.rb

Premium data is different from premium plan and I do not believe these will be deprecated anytime soon.

waruboy:
@aceunreal Ah, I think it is not applicable for google places.
From: https://developers.google.com/places/web-service/search

> Note for Google Maps APIs Premium Plan customers: You must include an API key in your requests. You should not include a client or signature parameter with your requests.

Or is this different with what you are talking?

aceunreal:
@waruboy 👍  Looks like client/signature is not supported for Google places

waruboy:
Thanks :smile:

Issue #933, opened 2015-10-26
geocoder_spoofable_ip returns "ip_address:port"


ipepe:
geocoder_spoofable_ip on my jruby azure rails server returns "23.102.21.198:7063" and because of that it can't find proper match

ipepe:
For now I added to my geocoder initializer file:
```ruby
module Geocoder
  module Request
    alias_method :old_geocoder_spoofable_ip, :geocoder_spoofable_ip

    def geocoder_spoofable_ip
      ip = self.old_geocoder_spoofable_ip
      ip = ip.to_s.split(":").first if ip.include?(".") && ip.include?(":")
      ip
    end
  end
end
```

TrangPham:
Beginner task:

Add two new methods between line 56 and 57: 
```ruby
addrs = geocoder_remove_port_from_addresses(addrs)
addrs = geocoder_reject_non_ipv4_addresses(addrs)
```
https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/request.rb#L57

PR #1150, opened 2017-02-13
933 remove port from geocoder_spoofable_ip

ghbooth12:
Hello. This PR is about the issue #933. 
I created `geocoder_remove_port_from_addresses(ip_addresses)` and `geocoder_reject_non_ipv4_addresses(ip_addresses)` to insure that the `ip_addresses` contains the valid ip address without port number. Then these two method are called in `geocoder_spoofable_ip` method.
I added tests for the cases when the `ip_addresses` includes port numbers and invalid ip addresses.


TrangPham:
This looks good to me!

PR #1153, opened 2017-02-25
Update smarty streets search results coordinates method to return nil 

RoyLee:
Currently when running a search via `Geocoder.coordinates("a zip code")` for certain zip codes (eg 96628) that exist, but for some reason doesn't exist in the Smarty Streets database, the coordinates method returns `[nil,nil]` which is not very useful. What would be more useful is if the method returns an actual nil value. These changes enable that.


alexreisner:
Thanks! This looks good. A more succinct way of writing the condition might be `result.compact.empty?`. In case you're interested in that kind of thing.

RoyLee:
Great, I'll make the modifications based on your feedback, and combine the changes into one commit.

PR #1154, opened 2017-02-25
fix comment describing the required cache interface

asafdav2:
cache must respond to `#del`, not `#keys`


alexreisner:
Where do you see that `del` is required? (I agree that `keys` is not.)

asafdav2:
[here](https://github.com/alexreisner/geocoder/blob/master/README.md#caching)

alexreisner:
Ah, very good. Thanks!!

PR #1159, opened 2017-03-14
Add support for query params in mapbox lookup.

dounan:
The Mapbox  geocoding api supports additional parameters such as country. This PR enables geocoder to set these parameters for Mapbox lookups per query.

### DIFF
--- a/lib/geocoder/lookups/mapbox.rb
+++ b/lib/geocoder/lookups/mapbox.rb
@@ -9,7 +9,8 @@ module Geocoder::Lookup
     end
 
     def query_url(query)
-      "#{protocol}://api.mapbox.com/geocoding/v5/#{dataset}/#{url_query_string(query)}.json?access_token=#{configuration.api_key}"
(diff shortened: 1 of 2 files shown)


TrangPham:
Thanks @dounan for this PR!

This works fine as is and thanks for implementing a test too. Though it would be preferred if the look up implemented the params in the a similar fashion to other look ups. See this for example: https://github.com/alexreisner/geocoder/blob/6589b9b2cd6d1cdda06bd1f598fe9ad070d8e5a9/lib/geocoder/lookups/google_premier.rb

`url_query_string` method is used and `query_url_params` is overwritten to merge in the api key. This would also ensure that the access_key passed in for the specific query is used and then defaults to the configuration if it isn't passed in.

dounan:
The previous implementation overrode the url_query_string method in an incompatible way with query_url_params. Did a small refactor to get it to the same pattern as the other files.

TrangPham:
Thanks - and sorry that it took me a while to get back to you and merge this!

dounan:
No worries!

PR #1165, opened 2017-04-21
Add store#keys to Caching object requirements

ttilberg:
A `NoMethodError` exception was raised when I tried expiring the cache while using a PORO Cache <-> DB adapter that did not implement `#keys`.

As an aside, sorry if it was inappropriate to submit the PR without creating an issue first, but it seemed unnecessary to create an issue for such a small change solely to documentation. I'm interested if you feel the same way, as I'm relatively new to contributing in the wild.

### DIFF
--- a/README.md
+++ b/README.md
@@ -883,8 +883,8 @@ This example uses Redis, but the cache store can be any object that supports the
 
 * `store#[](key)` or `#get` or `#read` - retrieves a value
 * `store#[]=(key, value)` or `#set` or `#write` - stores a value
-* `store#keys` - retrieves array of existing keys
 * `store#del(url)` - deletes a value
+* `store#keys` - (Optional) Returns array of keys. Used if you wish to expire the entire cache (see below).
 
 Even a plain Ruby hash will work, though it's not a great choice (cleared out when app is restarted, not shared between app instances, etc).
 
--- a/README.md
+++ b/README.md
@@ -883,6 +883,7 @@ This example uses Redis, but the cache store can be any object that supports the
 
 * `store#[](key)` or `#get` or `#read` - retrieves a value
 * `store#[]=(key, value)` or `#set` or `#write` - stores a value
+* `store#keys` - retrieves array of existing keys
 * `store#del(url)` - deletes a value
 
 Even a plain Ruby hash will work, though it's not a great choice (cleared out when app is restarted, not shared between app instances, etc).
--- a/lib/geocoder/cache.rb
+++ b/lib/geocoder/cache.rb
@@ -40,6 +40,7 @@ module Geocoder
     #
     def expire(url)
       if url == :all
+        raise(NoMethodError, "The cache store must implement `#keys` when using `:all`") unless store.respond_to?(:keys)
         urls.each{ |u| expire(u) }
       else
         expire_single_url(url)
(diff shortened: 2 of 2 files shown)


alexreisner:
Thanks for this! (And no, an issue is not necessary.)

So, `#keys` is not currently listed as required because it's only needed for expiring all keys (not individual keys), which, at the time I wrote the docs, didn't seem like core cache functionality. Whether it is or isn't, I agree that it should be mentioned. What do you think about listing it as optional, and/or mentioning that it's only needed for expiring the whole cache?

ttilberg:
This makes me wonder if it were optional, there should be some code that wraps the cache expire methods, checking first if `keys` exists, so it can fail gracefully. 

But that also may just be covering an error up that you would need to see, in order to know something is amiss. I feel like since the initial objects expect the cache store to respond to #keys, it should be required. And, thankfully, it's probably no more troublesome to implement than the get method anyway. 

I'm not sure what the best direction is, and right now I'm only thinking in the context of a rdbs adapter, where `select keys` is no biggie.

alexreisner:
It is actually quite troublesome to implement in a system like memcached which cannot list the keys. That's why it's not required. If anything, I think `expire(:all)` should fail with a more useful error message. But clear(er) documentation would also be sufficient, for now.

ttilberg:
Okay, I better understand now and agree that it shouldn't be marked as required. In the area describing the anatomy of a caching object, perhaps have it be notated as optional with a short description as to when it's used, and perhaps just a quick reminder that your caching object should respond to `#keys` in the section that talks about how to expire `:all`... "If your cache object responds to `#keys` you can expire `:all` keys..." or some such? Too much blah blah? 

Feel free to add the changes to the docs as you best see fit, or I can add another commit tomorrow after getting more feedback from you to confirm this direction. 

Thanks!

ttilberg:
I've tried to keep the documentation simple while still getting the information across. I've also added information at runtime if you try expiring `:all` without implementing `:keys`.

I wasn't sure if the best place was right before the call for `#keys` or immediately after calling `expire(:all)`. I felt `:all` made more sense, so the error message could be more helpful, and specific to calling `:all`.

Let me know if you'd like this further improved. Thanks!

alexreisner:
This looks good, thanks. Just two things:

1. Because error messages can come from deep within code you didn't write, it'd be helpful to clarify where the exception comes from. So maybe a message like:

    "The Geocoder cache store must implement `#keys` in order for `expire(:all)` to work"

2. I think the logic would be clearer if there was a multi-line if/else/end structure: if `keys` is implemented, call it, else raise an exception. Make sense?

Other than that, this looks good to me.

ttilberg:
Okay, requested updates are pushed up.

Can I ask you -- given this back and forth and minor text fixes across three commits -- what is the typically desired submission process? It's nice to have the individual commits and the associated discussion between the iterations, but for the actual project, it should obviously be squashed.

Do contributors typically squash the series of commits after these resolutions and submit a new PR? Or force push to the branch? Or just let the maintainer deal with it? Thanks for the feedback.

alexreisner:
Thanks! Looks great. As for the process (it's so nice that you asked), I don't have a standard way of handling it. Some people rebase as they go and others force push new commits. If there's something substantive to discuss about the commits themselves I'll usually start a discussion about it, but usually I just squash/merge (and lightly edit the commit message). I think that makes the most sense in this case. Thanks again!

PR #1167, opened 2017-05-03
Add street and street number to Here's result 

alexshenia:
Here Map returns street and street number in result. I added two methods (like a google lookup result) `route` and `street_number`.


VILATTI:
👍

atiroch:
👍

mic-2000:
👍

PR #1169, opened 2017-05-11
Allow to limit the objects that will be geocoded by rake task

fleveque:
That's a proposal to limit the objects that will be geocoded by the rake task, using an environment variable (LIMIT).

It could be useful if you want to geocode thousands of objects and don't want to reach the per-day limits of the geocoding services (for example if you are running the rake task in production and you are using the service for other porpoises.

It won't use batches when LIMIT is used, because _limit_ and _find_each_ don't work together.

Comments and/or corrections are very welcome :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,6 +114,10 @@ Geocoder will print warnings if you exceed the rate limit for your geocoding ser
 
     rake geocode:all CLASS=YourModel SLEEP=0.25 BATCH=100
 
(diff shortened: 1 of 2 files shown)


TrangPham:
This looks good to me. @alexreisner ?

alexreisner:
Thanks! Sorry for the delay.

fleveque:
You are welcome! Thanks for merging it 👍

PR #1175, opened 2017-05-16
Use https for freegeoip

mehanoid:
Requests to freegeoip.io now works only with https. So changed protocol for default :host, while continuing to use http for custom :host and allowing to set protocol with :use_https for custom :host

### DIFF
--- a/README.md
+++ b/README.md
@@ -737,7 +737,7 @@ This uses the PostcodeAnywhere UK Geocode service, this will geocode any string 
 * **Documentation**: http://github.com/fiorix/freegeoip/blob/master/README.md
 * **Terms of Service**: ?
 * **Limitations**: ?
-* **Notes**: If you are [running your own local instance of the FreeGeoIP service](https://github.com/fiorix/freegeoip) you can configure the host like this: `Geocoder.configure(freegeoip: {host: "..."})`.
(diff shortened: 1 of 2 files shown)


alexreisner:
This looks good, however I don't think we should be looking at `configuration[:use_https]` to determine what protocols are supported. Is there any problem with, when a custom host is specified, returning `[:https, :http]`?

mehanoid:
@alexreisner, you are right, we do not neet to check this, and result will be the same. I've updated PR.

alexreisner:
@mehanoid thanks! Only other thing is: should we mention `use_https: false` in the README? That change seems potentially misleading since it's not required with a custom host.

mehanoid:
I think we should mention that this option exists, otherwise it's not obvious how to specify a host with https. Maybe it's worth changing the example to something like this:
```
Geocoder.configure(freegeoip: {
    host:      "...", # default: "freegeoip.io"
    use_https: true   # default: false
})
```

alexreisner:
I think it should be clear from the other places in the README where HTTPS is addressed. In this particular sentence we're talking about how to configure a host. I don't think the protocol is particularly relevant. I mean, it's not _irrelevant_, but I think it's extraneous (and potentially confusing) info in this context.

alexreisner:
Merged, minus README changes (which can be discussed later if desired--want to release new version now). Thank you!!

PR #1177, opened 2017-05-16
add ip_lookup option to initializer template

mcfiredrill:
I noticed this option was in the README, but not in the template, so it
might be nice to have it generated in the template as well.


alexreisner:
Thanks!

PR #1179, opened 2017-05-17
Comma in wrong position in readme :bing

vasspilka:


alexreisner:
Hmmm... You're right that the comma's location doesn't make sense, but I doubt Bing allows 500,000 queries daily when there's also a 125,000 yearly limit. We probably need to update the limits based on current Bing documentation.

TrangPham:
Looking at the bing documentation:
```
Phone: Can exceed 125,000 cumulative billable transactions per calendar year.
Public Windows app and Windows Phone apps: Can exceed 50,000 cumulative billable transactions within any 24-hour period.
```
So looks like it it's supposed to be `50,000`

PR #1187, opened 2017-05-29
Change documentation about pelias

piotrj:
It seems that pelias section had the same links as mapzen. Changed that so that it points to pelias github page.

### DIFF
--- a/README.md
+++ b/README.md
@@ -600,13 +600,13 @@ The [Google Places Search API](https://developers.google.com/places/web-service/
 
 #### Pelias (`:pelias`)
 
-* **API key**: required
+* **API key**: configurable (self-hosted service)
 * **Quota**: none (self-hosted service)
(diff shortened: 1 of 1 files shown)


alexreisner:
Thanks for this! You're right that this should be updated. However, I think we can/should be more specific on a couple of items:

* For API Key could we say something more specific, like "configurable"? I think the current text makes it sound like the API key is self-hosted, which doesn't make any sense.
* For Terms of Service we can link to [the data licenses page](https://github.com/pelias/pelias/blob/master/data_licenses.md).
* Limitations bullet should not be removed.

piotrj:
@alexreisner all your comments make sense. I was just reading through README and noticed that so I figured, why not send a PR :) Applied your comments.

alexreisner:
Looks great! Thank you!

Issue #1118, opened 2016-11-18
Pass query options through geocoded_by?


toppa:
I'm trying to figure out if it's possible to pass query options with `geocoded_by` - I don't see this mentioned in the docs. Specifically, I want to pass arguments to the Googe API's `components` parameter. I can do it with `Geocoder.search`, but I don't see how to do it with my method that's invoked by `geocoded_by` - it seems to want only a single string returned, and if I correctly understand its option for passing a block, it's for handling results. Is it possible to do this?

TrangPham:
Hi @toppa 

There are a few things that you could pass to `geocoded_by`. You can see those options here: https://github.com/alexreisner/geocoder/blob/6589b9b2cd6d1cdda06bd1f598fe9ad070d8e5a9/lib/geocoder/models/active_record.rb#L11-L22

But currently no way to handle passing params. We are open to accepting a pull request for this feature as this question has been asked before by others: https://github.com/alexreisner/geocoder/issues/765

Currently this can be achieved only by writing custom geocode method. See the linked issue for example. 

Thanks!

toppa:
Thanks for the quick reply. It looks like we would need to [open this up](6589b9b2cd6d1cdda06bd1f598fe9ad070d8e5a9/lib/geocoder/stores/base.rb#L93) to allow for user supplied params. This is my first time looking at the Geocoder internals, so I'm not sure offhand how `options[:user_address]` gets populated from the `:geocoded_by` method. But my thought is to maybe add support for something like `:geocoded_by_params` which would populate a new `options[:user_params]`, which we could add to that `query_options` hash. Does that general idea make sense?

TrangPham:
@toppa That is correct. `options[:user_params]` should be adding into the `query_options`

Instead of `:geocoded_by_params`, I would suggest just calling it `:params` to match the style of the search function. This would also be nicer on `reverse_geocoded_by`. So that calling would look like:

```ruby
geocoded_by :address, :params => {region: 'US'}

reverse_geocoded_by :lat, :lon, :params => {region: 'US'}
```

Here is where the `:user_address` is coming from: https://github.com/alexreisner/geocoder/blob/6589b9b2cd6d1cdda06bd1f598fe9ad070d8e5a9/lib/geocoder/models/active_record.rb#L14

[...]

PR #1188, opened 2017-06-02
Adding params for geocoded_by and reverse_geocoded_by

TrangPham:
For #1118 

Wondering if there's a better way to test this
Considering using `Mocha` gem to be able to check if method was called with certain values.


TrangPham:
Hmmm - not sure why the travis tests say failed on here - when I click the link it actually shows each test suite in the matrix passing. 

@alexreisner could you review my diff?

alexreisner:
Yes, sorry for the delay--this looks good to me!

Issue #1069, opened 2016-06-18
Search query not in English


MSprut:
Can I make Geocoder.search query not in English? Thanks.

TrangPham:
Yes - you can do this if the API you are querying supports this functionality. 

For example google api specifications for languages are here: https://developers.google.com/maps/faq#using-google-maps-apis

The way you'd use this with geocoder is to pass the params through:
```Ruby
Geocoder.search('123 rue fake', : language => 'fr')
```

I'll add this to the read me as well

MSprut:
Thanks TrangPham. I tried such method. This method changes language of a reply to the request, but not language of request.  I can be mistaken, but it didn't help me. I tried to send from rails console a request with the address in Russian and the geocoder returned "Invalid request". It turned out that it is necessary to convert the text of request in Russian into the utf8 format.

TrangPham:
Can you please paste your Geocoder config as well as the specific Geocoder.search you are sending? Also what version of Geocoder are you using?

Thanks

MSprut:
irb(main):003:0> Geocoder.config
=> {:timeout=>3, :lookup=>:google, :ip_lookup=>:freegeoip, :language=>:ru, :http_headers=>{}, :use_https=>false, :http_proxy=>nil, :https_proxy=>nil, :api_key=>nil, :cache=>nil, :cache_prefix=>"geocoder:", :basic_auth=>{}, :logger=>:kernel, :kernel_logger_level=>2, :always_raise=>[], :units=>:mi, :distances=>:linear}
---------------------------------------------------------------------------------------------
irb(main):004:0> Geocoder.search("18 Мичурина улица, Луганск")
Geocoding API error: 400 Bad Request
Google Geocoding API error: invalid request.
=> []
--------------------------------------------------------------
irb(main):006:0> Geocoder.search(("18 Мичурина улица, Луганск").encode("utf-8"))
[...]

Issue #609, opened 2014-02-24
Stop Using Deprecated Method: URI.escape


alexreisner:
This method is used in the [Bing lookup](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookups/bing.rb#L30). A common response to the problem is to replace `URI.escape` with `CGI.escape`. However, these methods do not behave the same way. To wit:

```
> URI.escape ' '  # => "%20"
> CGI.escape ' '  # => "+"

> URI.escape '+'  # => "+"
> CGI.escape '+'  # => "%2B"

> URI.escape ':'  # => ":"
> CGI.escape ':'  # => "%3A"
```

Switching to `CGI.escape` breaks the Bing lookup (no results are returned). Research needs to be done to understand this issue (and what Bing expects). For a starting point see [this good answer on Stack Overflow](http://stackoverflow.com/questions/2824126/whats-the-difference-between-uri-escape-and-cgi-escape#answer-13059657).

malandrina:
What is the status of this? It looks like `URI.escape` is available in the most recent version of Ruby ([reference](http://ruby-doc.org/stdlib-2.2.0.preview1/libdoc/uri/rdoc/URI/Escape.html)).

alexreisner:
It is available, but it was [marked as "obsolete" in Sep 2009](https://github.com/ruby/ruby/commit/238b979f1789f95262a267d8df6239806f2859cc) and [remains so today](https://github.com/ruby/ruby/blob/trunk/lib/uri/common.rb#L98). I'm not really sure what "obsolete" means in this context. Maybe it's different from "deprecated" but I'd like to at least get confirmation that it's fine to keep using `URI.escape` before closing this issue.

stevecj:
When I run the test suite using Ruby 2.3.1, I get several warnings saying "base_url': warning: URI.escape is obsolete". Some of the locations of those warnings are…

./lib/geocoder/lookups/bing.rb:34:in `base_url'
./lib/geocoder/lookups/maxmind_geoip2.rb:18:in `query_url'

Amith-catadoe:
> This method is used in the [Bing lookup](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookups/bing.rb#L30). A common response to the problem is to replace `URI.escape` with `CGI.escape`. However, these methods do not behave the same way. To wit:
> 
> ```
> > URI.escape ' '  # => "%20"
> > CGI.escape ' '  # => "+"
> 
[...]

PR #1189, opened 2017-06-03
force encode google query as utf-8

swentling87:
In reference to issue #1069

### DIFF
--- a/lib/geocoder/lookups/google.rb
+++ b/lib/geocoder/lookups/google.rb
@@ -22,7 +22,7 @@ module Geocoder::Lookup
     end
 
     def query_url(query)
-      "#{protocol}://maps.googleapis.com/maps/api/geocode/json?" + url_query_string(query)
+      "#{protocol}://maps.googleapis.com/maps/api/geocode/json?" + url_query_string(query).encode("utf-8")
     end
 
     private # ---------------------------------------------------------------


alexreisner:
Thanks for this! Two things:

1. Do we have confirmation (in the form of documentation) that Google expects/prefers UTF-8?
2. If so, then we should merge this, but will have to wait until Ruby 1.9.3 support is dropped (since strings don't have the `encode` method until 2.0.

swentling87:
I spent some time reading through the documentation and it looks like what the Google API is looking for is not necessarily UTF-8 but just proper URL-encoding. In their best practices section it does discuss however that UTF-8 encoding appears preferred. [https://developers.google.com/maps/documentation/geocoding/web-service-best-practices#BuildingURLs](url)

> All characters to be URL-encoded are encoded using a '%' character and a two-character hex value corresponding to their UTF-8 character

It is already encoding it properly for URL-encoding, but making it UTF-8 will help convert characters otherwise not accounted for to be used in the URL. Some of my search also led me to some info that may help with issue #609. In reading about URL encoding it seems like that issue could be fixed by how that request is encoded and there are a lot of options found here: [https://stackoverflow.com/questions/2824126/whats-the-difference-between-uri-escape-and-cgi-escape](url)

Notably, CGI is already used and may be a good switch from using URI.escape().

PR #1193, opened 2017-06-23
Added PickPoint provider.

cylon-v:


TrangPham:
Hi I am confused at why you're reusing the Nominatim lookup for this. Unless there is a specific association between pick point and nominatim, even though they might currently share a lot of the same format, I think you should be inheriting from the base lookup and base result instead. 

Thanks for this PR! :)

cylon-v:
Hi @TrangPham, PickPoint is a commercial service based on Nominatim. Basically it's exactly the same but has minor differences, i.e. in URL format.

TrangPham:
I can't find the documentation on the PickPoint site which says that. I'm okay with inheriting from nominatim if we know for sure that's what they are relaying on.

cylon-v:
I'm PickPoint developer :).

TrangPham:
:D Okay! thanks

cylon-v:
Thank you!

PR #1194, opened 2017-06-28
add #city_district, #state_district, #neighbourhood to Nominatim response

yurivm:
First of all, thanks a lot for this gem!
I've been using it for a while and recently switched to Nominatim. I noticed a few address parts were missing, namely:
- state_district
- neighbourhood
- city_district

Those are often present in Nominatim responses but not always, so I added a fixture and specs.

Thanks for checking!


TrangPham:
Looks good to me! Thank you for PR and adding tests!

PR #1196, opened 2017-07-01
add bounds detection for google results

jormon:
some google geocoding requests have bounds within the result.  add a
helper to extract these.

### DIFF
--- a/lib/geocoder/results/google.rb
+++ b/lib/geocoder/results/google.rb
@@ -120,19 +120,29 @@ module Geocoder::Result
     def precision
       geometry['location_type'] if geometry
     end
-    
+
     def partial_match
       @data['partial_match']
     end
-    
+
     def place_id
       @data['place_id']
-    end  
+    end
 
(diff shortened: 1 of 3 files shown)


TrangPham:
Hi thanks for the PR! You said 'some' of the responses have it. Could you
also ensure that there's a test cause for when the response doesn't have it?

jormon:
@TrangPham thanks for the thoughts re: testing -- I added that test case.  Note that this required a slight code change too since we now want divergent behavior from `#viewport` when there is no data.

alexreisner:
@jormon thanks for this, and for adding the test.

For anyone who (in addition to me) is baffled by the difference between "bounds" and "viewport," it's explained in Google's docs: https://developers.google.com/maps/documentation/geocoding/intro#Results

jormon:
@TrangPham will this get merged?

alexreisner:
@jormon sorry for the delay, and thanks!

PR #1207, opened 2017-07-24
Add time_zone to Opencagedata result

samoli:
The OpenCage API [may include a time zone in the response](https://geocoder.opencagedata.com/api#annotations). This PR adds `#time_zone` to the `Opencagedata` result so it can be accessed when it's available.

`Geocoder::Result#time_zone` has been previously implemented in  [`Maxmind`](https://github.com/alexreisner/geocoder/blob/6589b9b2cd6d1cdda06bd1f598fe9ad070d8e5a9/lib/geocoder/results/maxmind.rb#L49) and [`Pointpin`](https://github.com/alexreisner/geocoder/blob/6589b9b2cd6d1cdda06bd1f598fe9ad070d8e5a9/lib/geocoder/results/pointpin.rb#L31) so hopefully it makes sense to add it to `Opencagedata` too.

### DIFF
--- a/lib/geocoder/results/opencagedata.rb
+++ b/lib/geocoder/results/opencagedata.rb
@@ -75,6 +75,15 @@ module Geocoder::Result
       [south, west, north, east]
     end
 
+    def time_zone
+      # The OpenCage API documentation states that `annotations` is available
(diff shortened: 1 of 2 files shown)


alexreisner:
Looks great, thanks!

Issue #1208, opened 2017-07-26
No rails version specified for migration maxmind:geolite_city


ydakuka:
https://github.com/alexreisner/geocoder/blob/master/README.md#ip-address-local-database-services
When I generate a migration to create a table:
```
rails generate geocoder:maxmind:geolite_city
```
I get:
```
rails aborted!
StandardError: An error has occurred, this and all later migrations canceled:

Directly inheriting from ActiveRecord::Migration is not supported. Please specify the Rails release the migration was written for:

  class GeocoderMaxmindGeoliteCity < ActiveRecord::Migration[4.2]
```

### Expected behavior
class GeocoderMaxmindGeoliteCity < ActiveRecord::Migration[5.1]

### Actual behavior
class GeocoderMaxmindGeoliteCity < ActiveRecord::Migration

### Environment info

* Geocoder version: 
geocoder (1.4.3)
* Rails version: 
rails (5.1.2)

PR #1209, opened 2017-07-26
specify rails version for migrations

ydakuka:
In reference to issue https://github.com/alexreisner/geocoder/issues/1208


alexreisner:
Excellent! Thanks for this. Just a couple of minor things:

1. I think the name of the `rails5?` method should include an underscore (`rails_5?`)--any objections to changing that?
2. Can you move that method and the `migration_version` method somewhere else in the hierarchy so they're not duplicated?

ydakuka:
@alexreisner fixed it.

alexreisner:
Thank you!

PR #1214, opened 2017-08-23
add support for sequel ORM

fran-worley:
Some of us (or maybe it's just me!) would love to be able to use the powers of this gem outside of rails. 

So for those that are interested, here is a first stab at adding support for Jeremy Evans Sequel ORM.

There are alot of improvements that could be made, but I wont bother you with those if there is no interest from the wider community!


fran-worley:
Any ideas on why these fail? I can't see what I've changed to upset those tests...

`Geocoder::ConfigurationError: When using MaxMind you MUST specify a service name: Geocoder.configure(:maxmind => {:service => ...}), where '...' is one of: [:country, :city, :city_isp_org, :omni]`

test/test_helper.rb
```diff
@@ -7,23 +7,39 @@
...
+    require 'active_record'
+
+    # Establish a database connection
+    ActiveRecord::Base.configurations = configs
+    ActiveRecord::Base.establish_connection(db_name.to_sym)
+    ActiveRecord::Base.default_timezone = :utc
+
+    ActiveRecord::Migrator.migrate('test/db/migrate', nil)
+  elsif ENV['ORM'] == 'sequel'
+    require 'sequel'
```
TrangPham: this fails on Travis :(

fran-worley:
Hi @TrangPam as I said, this is a WIP!! I just want going to bother cleaning it up any further if you guys weren't interested. 

If you think Sequel would be a good ORM to support then I'll take this a bit further, if not I'll release it separately.

TrangPham:
Thanks @fran-worley. We'd love to support another ORM within geocoder, and would require help maintaining it from someone who uses the ORM regularly. 

If you want to see if there is more interest in this, then we can wait for others to comment here or on issues. 

We are working on updating our travis matrix and fixing the dependancies in our tests. See: #1224

alexreisner:
Hey, sorry for the delay on this. @fran-worley thanks so much for thinking of this and getting started on it. In principle I love the idea of supporting Sequel. In practice, supporting another ORM can be a lot of work. We ended up mostly dropping MongoDB support because of the difficulty of automating tests and such. But since Sequel doesn't require installation of any new DB systems, it would be much easier than Mongo. I'd be up for adding Sequel support but labeling it "EXPERIMENTAL" and not making a long-term commitment until we see how it goes in the wild (all kinds of weird stuff pops up when thousands of people are using your code).

A couple of things:

1. I share @TrangPham 's confusion about [this code comment](https://github.com/alexreisner/geocoder/pull/1214/files#r139224994).
2. It looks like there's a lot of code duplication. I wonder if what you're saying with that comment is that if we reorganized the ActiveRecord store, the Sequel store could inherit from it and we could reduce the duplication. Refactoring the ActiveRecord store is a lot of work, and makes me a little nervous since that code is so fragile. But that's also a reason to fix it.

fran-worley:
@alexreisner @TrangPham - Thanks for coming back to me. 

Having dug through both the code base, AR is pretty tightly baked into this gem which makes it harder to cleanly add other ORMs (and test them). So yes, you are exactly right, my comment was due the code duplication caused by this. 
Essentially, I didn't want to muck about the the AR code too much, but with that goal in mind I ended up with a lot of duplication.

I'll take another look at this and see whether I can come up with anything worthwhile.

alexreisner:
Closing, at least for now. As @fran-worley rightly points out, the gem is too tightly coupled to ActiveRecord. If that were to be fixed then we should reconsider this. But for now I'm calling a moratorium on any new ORMs (see also #1248).

PR #1216, opened 2017-08-29
Update LocationIQ geocoding provider with limits and handle errors

SrihariThalla:
Disclosure: I work at LocationIQ (Unwired Labs). This PR is sponsored by Unwired Labs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -498,13 +498,13 @@ The [Google Places Search API](https://developers.google.com/places/web-service/
 #### LocationIQ (`:location_iq`)
 
 * **API key**: required
-* **Quota**: 6 request/second (30k req/day), then ability to purchase more
+* **Quota**: 60 requests/minute (2 req/day, 10k req/day), then ability to purchase more
 * **Region**: world
 * **SSL support**: yes
-* **Languages**: ?
-* **Documentation**: http://locationiq.org/#docs
+* **Languages**: worldwide
+* **Documentation**: https://locationiq.org/#docs
 * **Terms of Service**: https://unwiredlabs.com/tos
-* **Limitations**: [Data licensed under Open Database License (ODbL) (you must provide attribution).](http://www.openstreetmap.org/copyright)
+* **Limitations**: [Data licensed under Open Database License (ODbL) (you must provide attribution).](https://www.openstreetmap.org/copyright)
 
 #### OpenCageData (`:opencagedata`)
 
--- a/README.md
+++ b/README.md
@@ -498,10 +498,10 @@ The [Google Places Search API](https://developers.google.com/places/web-service/
 #### LocationIQ (`:location_iq`)
 
 * **API key**: required
-* **Quota**: 60 requests/minute (2 req/day, 10k req/day), then [ability to purchase more](http://locationiq.org/#pricing)
+* **Quota**: 60 requests/minute (2 req/sec, 10k req/day), then [ability to purchase more](http://locationiq.org/#pricing)
 * **Region**: world
 * **SSL support**: yes
-* **Languages**: worldwide
+* **Languages**: ?
 * **Documentation**: https://locationiq.org/#docs
 * **Terms of Service**: https://unwiredlabs.com/tos
 * **Limitations**: [Data licensed under Open Database License (ODbL) (you must provide attribution).](https://www.openstreetmap.org/copyright)
--- a/lib/geocoder/lookups/location_iq.rb
+++ b/lib/geocoder/lookups/location_iq.rb
@@ -32,9 +32,9 @@ module Geocoder::Lookup
         when "Key not active - Please write to contact@unwiredlabs.com"
           raise_error(Geocoder::RequestDenied, doc['error'])
         when "Rate Limited"
-          raise_error(Geocoder::OverQueryLimitError, doc['error')
+          raise_error(Geocoder::OverQueryLimitError, doc['error'])
         when "Unknown error - Please try again after some time"
-          raise_error(Geocoder::InvalidRequest, doc['error')
+          raise_error(Geocoder::InvalidRequest, doc['error'])
         end
       end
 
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -498,13 +498,13 @@ The [Google Places Search API](https://developers.google.com/places/web-service/
 #### LocationIQ (`:location_iq`)
 
 * **API key**: required
-* **Quota**: 6 request/second (30k req/day), then ability to purchase more
+* **Quota**: 60 requests/minute (2 req/day, 10k req/day), then ability to purchase more
```
TrangPham: What's the 2 req/day for? Documentation seems to inidicate just 10K req/day.

lib/geocoder/lookups/location_iq.rb
```diff
@@ -7,23 +7,35 @@ def name
       "LocationIq"
     end
 
+    def use_ssl?
+      true
```
TrangPham: The documentation examples on documentation link seems to only show http examples. Are you sure that https works?

lib/geocoder/lookups/location_iq.rb
```diff
@@ -7,23 +7,35 @@ def name
...
+    end
+
     def required_api_key_parts
       ["api_key"]
     end
     
     def query_url(query)
-      method = query.reverse_geocode? ? "reverse.php" : "search.php"
-      host = configuration[:host] || "locationiq.org/v1"
```
TrangPham: Has the ability to host your own version been removed? Should we still handle that ability since those with local copies would still have them?

README.md
SrihariThalla: We clarified the rate limits [under FAQs](https://locationiq.org/#faqs)

lib/geocoder/lookups/location_iq.rb
SrihariThalla: Yes, `https` works along with `http`

lib/geocoder/lookups/location_iq.rb
SrihariThalla: [Nominatim](http://nominatim.org/) is the one to host and not [LocationIQ](https://locationiq.org/). I see that the host param is implemented https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookups/nominatim.rb#L17

However, a drawback with `Nominatim` provider is that it doesn't accept an API key. If `LocationIq` provider is allowed to have `configuration[:host]`, then hosting Nominatim behind an API key is possible. However, some error responses might be different as we have mutated them and added some other info (like our email address.)

Let me know your thoughts on this.

SrihariThalla:
Hi, @TrangPham!

Yes, I work for Unwired Labs / LocationIQ. Error responses are missing and we will update them soon possible.

Let me know your thoughts on using `LocationIq` provider for self-hosted versions and the resulting breaking changes if the `:host` is removed from it!

Thanks for your feedback!

lib/geocoder/lookups/location_iq.rb
TrangPham: I think it would be fine to remove. I will point those who have errors due to this to use the nominatim lookup instead of this one

lib/geocoder/lookups/location_iq.rb
TrangPham: If it works with both, should we let the user decide which one to use? It's available as a config setting in geocoder

TrangPham:
Also your tests have failed. :) what you'll have to do is create the json file for the error responses and then ensure those are returned in the mocks. Let me know if you need an example of this.

SrihariThalla:
@TrangPham All the tests are now passing! Let me know if I need to update anything else 😉

README.md
```diff
@@ -498,13 +498,13 @@ The [Google Places Search API](https://developers.google.com/places/web-service/
 #### LocationIQ (`:location_iq`)
 
 * **API key**: required
-* **Quota**: 6 request/second (30k req/day), then [ability to purchase more](http://locationiq.org/#pricing)
+* **Quota**: 60 requests/minute (2 req/day, 10k req/day), then [ability to purchase more](http://locationiq.org/#pricing)
```
alexreisner: I think there's a typo here (req/sec instead of req/day?)

README.md
```diff
@@ -498,13 +498,13 @@ The [Google Places Search API](https://developers.google.com/places/web-service/
...
 * **Region**: world
 * **SSL support**: yes
-* **Languages**: ?
-* **Documentation**: http://locationiq.org/#docs
+* **Languages**: worldwide
```
alexreisner: Does this mean every language on earth is supported? Is it possible to be more specific? This can be critical info when it comes to deciding on an API.

SrihariThalla: LocationIQ is a Nominatim (OpenStreetMap) service. You can safely assume it's a mirror. OpenStreetMap certainly doesn't support every language on Earth unless contributors add them!

I'll revert `worldwide` back to `?`.

README.md
SrihariThalla: Corrected ;)

lib/geocoder/lookups/location_iq.rb
SrihariThalla: @TrangPham @alexreisner Shall I update this to `["use_ssl"]` and let the users decide?

alexreisner: @SrihariThalla definitely. Thanks!

lib/geocoder/lookups/location_iq.rb
```diff
@@ -8,7 +8,7 @@ def name
     end
 
     def use_ssl?
-      true
+      ["use_ssl"]
```
alexreisner: Sorry, I misspoke here. We actually shouldn't be implementing `use_ssl?` in lookup subclasses. `supported_protocols` should be used instead. But since the default is HTTP and HTTPS, this method can simply be deleted.

SrihariThalla: Removed `use_ssl?`

SrihariThalla:
All the tests are now passed. Let me know if anything has to be updated!

TrangPham:
Thanks for updating~!

PR #1225, opened 2017-09-26
Removing ruby versions that are not supported anymore from our travis…

TrangPham:
… matrix

```
WARNING: Please be aware that you just installed a ruby that is no longer maintained (2014-02-23), for a list of maintained rubies visit:

    http://bugs.ruby-lang.org/projects/ruby/wiki/ReleaseEngineering

Please consider upgrading to ruby-2.2.1 which will have all of the latest security patches.
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: ruby
 sudo: false
 env:
-  global: 
+  global:
     - JRUBY_OPTS=--2.0
   matrix:
     - DB=
@@ -10,8 +10,6 @@ env:
     - DB=postgres
     - DB=mysql
 rvm:
-  - 1.9.3
-  - 2.0.0
   - 2.1.2
   - 2.2.2
   - 2.3.0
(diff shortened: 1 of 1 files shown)


TrangPham:
Also removing 2.1.2:

https://www.ruby-lang.org/en/news/2017/04/01/support-of-ruby-2-1-has-ended/

alexreisner:
Hmm...I don't think we should do this quite so suddenly. "Unsupported" Ruby versions aren't dead, they're just not receiving security updates. They may still be in (wide) use, and removing versions from the test matrix is tantamount to dropping support for them, which requires a minor version bump, and so should probably be merged into a separate branch (eg, `1.5.0`).

I think we can drop 1.9.3 support in the next minor release, and I'm open to arguments for dropping 2.0.x support as well, although I'm a little less comfortable with that, given that compatibility and ease of use are core Geocoder features. For the others, we should introduce version-specific gemfiles and select compatible versions of gems (eg, `byebug`).

And, when we do drop 1.9.3 support, we can also remove `gemfiles/Gemfile.ruby1.9.3`. Yay!

PR #1230, opened 2017-10-10
[db-ip] Handle Geocoder::OverQueryLimitError gracefully

cv:
`Geocoder::OverQueryLimitError` errors were not being generated correctly, and the response was coming back empty but with no indication.


TrangPham:
Looks good to me - could you add another test case and fixture for it, similar to: https://github.com/alexreisner/geocoder/blob/master/test/fixtures/db_ip_com_invalid_key

Thanks!

cv:
Thanks for having a look at this @TrangPham! I've added the fixture file, but not quite sure how it gets picked up by the tests. Got a pointer?

TrangPham:
@cv Thanks for your quick response :)

Here's an example: https://github.com/alexreisner/geocoder/blob/6589b9b2cd6d1cdda06bd1f598fe9ad070d8e5a9/test/unit/lookups/geocodio_test.rb#L63-L67

cv:
Looks like there's an unrelated CI failure with jruby 1.9, but other than that all tests are passing now! :)

TrangPham:
Thanks! Looks like the error is unrelated :)

PR #993, opened 2016-03-02
Update README.md

sadfuzzy:
Syntax highlighting, Single code style, Copyright years updated (2016)

PR #1234, opened 2017-10-14
Update README.md

cle61:


alexreisner:
Please see #993. Basically, the code examples are so simple that I don't think it's worth sacrificing the readability of the plain-text file for some syntax highlighting.

PR #1223, opened 2017-09-25
Add support for DB-IP.com geocoding service, take 2

cv:
This PR is a continuation of https://github.com/alexreisner/geocoder/pull/1063, which seems to have been abandoned in 2016, and I'd like to help see it through to a release.

PR #1235, opened 2017-10-17
Add support for ipsidekick.com geocoding service

hboon:


alexreisner:
Thanks for this! The tests are failing, however, primarily because of some missing JSON responses. Could you take a look at that? For recent examples, please see #1223 and #1193.

hboon:
Ah. My bad for forgetting about the tests.

Question though. The geocoding service doesn't support lat-long (yet. I'm the developer). But hear me out, this is a sample output:

```
{
  "ip":"17.142.160.59",
  "country": {
    "code":"US",
    "name":"United States"
  },
  "currency" : {
    "code" : "USD",
    "decimals" : 2,
    "name" : "US dollar"
   },
  "timezone": {
    "gmtOffset":"GMT-7:00",
    "name":"America/Los_Angeles"
  },
  "holiday":"Independence Day"
}
```

If you'd rather all geocoders that work with the gem produce a lat-long output, I can understand that. Otherwise, I can have create a separate PR to discuss how to modify the test to reflect it not support all the fields in Ah. My bad for forgetting about the tests.

Question though. The geocoding service doesn't support lat-long (yet. I'm the developer). But hear me out, this is a sample output:

```
{
  "ip":"17.142.160.59",
  "country": {
    "code":"US",
    "name":"United States"
  },
  "currency" : {
    "code" : "USD",
    "decimals" : 2,
    "name" : "US dollar"
   },
  "timezone": {
    "gmtOffset":"GMT-7:00",
    "name":"America/Los_Angeles"
  },
  "holiday":"Independence Day"
}
```

If you'd rather all geocoders that work with the gem produce a lat-long output, I can understand that. Otherwise, I can have create a separate PR to discuss how to modify the test to reflect it not support all the fields in https://github.com/alexreisner/geocoder/blob/master/test/unit/result_test.rb#L99

alexreisner:
Hmm...interesting question. On one hand, none of the supported IP geolocation services produce really meaningful lat/lon coordinates because, at best, they're returning the location of the ISP, not the user's device (and often they only return the coords at the center of the user's country). However, developers using Geocoder expect to get lat/lon (often for purposes of doing distance calculations) so a lack of coordinates may be confusing (and could be the cause of a "bad experience" with your API). Do you have plans to add lat/lon coordinates in the future?

hboon:
> Do you have plans to add lat/lon coordinates in the future?

No definite plans to. I have some data, but like you say, they are often not accurate enough or become outdated quickly. Product-wise, I'm growing capabilities by attaching additional types of data — now currency, timeZone, holidays, hopefully more sooner.

It returns the country now which I know people have some use for it, but it's your call :)

alexreisner:
So, after thinking about this, I don't think it's a good idea to add a service that breaks such a fundamental expectation. If you add lat/lon in the future (even with the low degree of precision offered by other APIs), feel free to re-open this PR.

PR #1236, opened 2017-10-18
Update docs to include additional Mapbox options

rdlugosz:
Support for additional query params was added with commit
844da1c11c2e74777ee8aab40cf82cee0f68dfa8 but this was not reflected in
the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,7 +425,7 @@ The following is a comparison of the supported geocoding APIs. The "Limitations"
 * **Region**: world
 * **SSL support**: yes (required if key is used)
 * **Languages**: see https://developers.google.com/maps/faq#languagesupport
-* **Extra options**:
+* **Extra params**:
   * `:bounds` - pass SW and NE coordinates as an array of two arrays to bias results towards a viewport
   * `:google_place_id` - pass `true` if search query is a Google Place ID
 * **Documentation**: https://developers.google.com/maps/documentation/geocoding/intro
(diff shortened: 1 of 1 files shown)


alexreisner:
Thanks for this! Just some thoughts on formatting...

* The way to pass these extra options is described elsewhere in the readme, so I don't think we need the example here.
* These should probably be at a single-indentation-level, like the Google section--the "Common options include..." isn't necessary.
* They should probably be called "extra params" rather than "options." Seems like we should change the header in the Google section as well, unless you feel otherwise.

This part of the readme is so dense already--I'm just trying to keep it as readable as possible!

rdlugosz:
It's funny you should mention how much is already in the Readme... I was prompted to make this PR after it took me a while to figure out how to pass this stuff to Mapbox via Geocoder! :)

I trimmed things a bit per your suggestions & also switch the word "options" to "params" on both the Mapbox and Google sections.

Agree that it'd be best to not expand the readme too far, but also need to balance ease of finding examples of common/tricky things. Your call on whether it makes sense to include or not. Maybe long term the documentation could be broken out into separate files with expanded illustration of each geocoding endpoint? (Not signing up for that btw, but might be something that someone else wants to take on.)

alexreisner:
Thanks! And yes, I completely agree about the length of the README. We're going to do something about it at some point...not a quick or easy fix, as you noticed.

PR #1244, opened 2017-11-28
Compress cache values larger than 1 Kilobyte

lipanski:
We've been struggling with containing our geocoder cache under decent limits as some of the lookup responses tend to be very verbose. I recently discovered that Rails is actually [compressing cache values](https://github.com/rails/rails/blob/master/activesupport/lib/active_support/cache.rb#L739) beyond a certain size and has been doing so for quite some time (they've recently lowered the threshold from 16Kb to 1Kb).

So I thought that this would be a great improvement to geocoder. Compression is fairly cheap nowadays, especially because we're not dealing with huge documents here. And the fact that most (all?) of the lookups are delivering JSON helps quite a lot.

My implementation is inspired by the Rails implementation, with some simplifications:

- I'm not *marhsaling* anything because we're only dealing with Strings here.
- The problem of figuring out if a value was compressed or not: Rails solves this by wrapping the cached value inside an object (`Entry`) which they *marshal* again, but I've decided to just prepend compressed values with that *marker* thingy. I've done so for simplicity (it's also faster and easier to read when looking inside your cache store), but it comes with the downside of this bit more rigid structure.

I left the option *disabled* by default.

I'd love to see this a part of geocoder. Let me know what you think :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -943,6 +943,12 @@ You can also set a custom prefix to be used for cache keys:
 
 By default the prefix is `geocoder:`
 
+In some situations it helps a lot to **compress the cache values**. Geocoder will do that for values larger than 1 Kilobyte, provided that you enable cache compression inside your configuration:
+
+```ruby
+Geocoder.configure(:cache_compress => true)
+```
+
 If you need to expire cached content:
 
     Geocoder::Lookup.get(Geocoder.config[:lookup]).cache.expire(:all)  # expire cached results for current Lookup
--- a/lib/geocoder/cache.rb
+++ b/lib/geocoder/cache.rb
@@ -1,16 +1,22 @@
+require "zlib"
+
 module Geocoder
   class Cache
 
-    def initialize(store, prefix)
+    COMPRESS_THRESHOLD = 1024 # bytes
+    COMPRESS_MARKER = "geocoder/compressed;"
+
+    def initialize(store, prefix, options = {})
       @store = store
       @prefix = prefix
+      @compress = options.fetch(:compress, false)
     end
 
     ##
     # Read from the Cache.
     #
     def [](url)
-      interpret case
+      value = interpret case
         when store.respond_to?(:[])
           store[key_for(url)]
         when store.respond_to?(:get)
@@ -18,12 +24,22 @@ module Geocoder
         when store.respond_to?(:read)
           store.read key_for(url)
       end
+
+      if compressed?(value)
+        uncompress(value)
+      else
+        value
+      end
     end
 
     ##
     # Write to the Cache.
     #
     def []=(url, value)
+      if should_compress?(value)
+        value = compress(value)
+      end
+
       case
         when store.respond_to?(:[]=)
           store[key_for(url)] = value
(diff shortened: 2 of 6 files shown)


lipanski:
I compared the bytesize on the Madison Square Garden fixtures:

```
google / old: 1433 / compressed: 445
mapbox / old: 7311 / compressed: 1244
bing / old: 1556 / compressed: 749
```

...and the performance penalty on the mapbox example, cause it's the longest one:

```ruby
require "benchmark"
require "geocoder/cache"

compressed_cache = Geocoder::Cache.new({}, "", compress: true)
old_cache = Geocoder::Cache.new({}, "", compress: false)

content = File.read("test/fixtures/mapbox_madison_square_garden")
n = 5000

Benchmark.bm do |x|
  x.report(:write_compressed) { n.times { compressed_cache["key"] = content } }
  x.report(:write_old) { n.times { old_cache["key"] = content } }
end

Benchmark.bm do |x|
  x.report(:read_compressed) { n.times { compressed_cache["key"] } }
  x.report(:read_old) { n.times { old_cache["key"] } }
end
```

```
       user     system      total        real
write_compressed  0.450000   0.010000   0.460000 (  0.461400)
write_old  0.010000   0.000000   0.010000 (  0.006051)
       user     system      total        real
read_compressed  0.100000   0.000000   0.100000 (  0.096833)
read_old  0.000000   0.000000   0.000000 (  0.004626)
```

mileage may vary but I think it's a good trade-off.

lipanski:
not exactly sure what to do about the jruby failures in Travis (seems not related to my changes).

alexreisner:
Thanks! I like this idea!

I assume the reasoning behind the `COMPRESS_THRESHOLD` is that below a certain point the memory saved isn't worth the speed sacrificed. But why 1KB? This isn't critical, but as long as we're thinking about it, how might we choose the optimal threshold? It seems like it could potentially be configurable, and might often be enabled on a per-lookup basis (e.g., to deal with Mapbox's large responses), but maybe there's something about the nature of the data received from any lookup that makes one particular value (eg, 1KB) optimal, in which case I would not want to make it configurable.

lipanski:
@alexreisner thanks. as said, I picked up the value from [Rails](https://github.com/rails/rails/pull/31147) - though that PR doesn't explain the choice either. there's definitly a threshold under which it's not worth compressing any more, probably because the compression rate drops towards the lower boundry and you're not saving much cache memory/storage.

the JSON format helps, because it's more compression-friendly than other formats. so, as long as most (all?) lookups respond with JSON it's not worth treating them distinctly.

I'd keep things simple, but I can also expose this as an option (with a default value).

let me know what you think.

lipanski:
ping @alexreisner :)

lime:
Great work!

This is a bit tangential, but I thought I should bring Moneta to your attention: 

https://github.com/minad/moneta
http://www.rubydoc.info/gems/moneta/

It lets you construct a cache store by combining the exact pieces you want your caching strategy to consist of (think middlewares). Then you can use that as the cache in e.g. Rails or Geocoder. For example, replicating this PR + using Redis for storage would look something like the following:

```ruby
geocoder_cache = Moneta.build do
  use :Transformer, value: :zlib
  adapter :Redis
end

Geocoder.configure(cache: geocoder_cache)
```

The above doesn't include the 1KB limit (I'm not sure if there's an easy way to make Moneta transformers conditional), but should otherwise be equivalent.

This makes it easy to add caching features specifically where you need them, such as expiration (`:Expires`) or an intermediate in-memory LRU cache (`:Stack` + `:LRUHash`), without the gem developer having to include those features in their default cache.

Just thought this could be useful. :)

lipanski:
@lime cool, didn't know moneta could do that. do you know if its smart enough to handle cases where only part of the keys have been compressed? (e.g. introducing compression for future entries but keeping the old entries uncompressed and still readable)

lime:
I don't know, but I suspect that it wouldn't work. You could probably set up a transformer that falls back to the raw value if decompression fails, but I don't think it does that out of the box. It would be interesting to see if you could build a store with some sort of branching, the way you've done here with a marker. 🤔

In general, I would think that any changes to your transformers should be accompanied by a cache bust (e.g. by incrementing a version in the prefix) unless it's a _really_ expensive operation.

alexreisner:
@lime thanks for mentioning this. @lipanski the reason I'm hesitating to merge is actually very much related. I'm not yet convinced the cache size should be managed by an application which uses a cache, as opposed to the cache itself.

alexreisner:
Closing, because I think this adds complexity that should be the responsibility of the cache, but this was a useful discussion (and good code!). Thanks, and sorry.

PR #1252, opened 2018-01-02
Handle IPv6 addresses when removing port numbers in request.location. Closes #1210

### LINKED ISSUE #1210: IPv6 Causes IPAddr::InvalidAddressError
The [`Geocoder::Request#geocoder_remove_port_from_addresses` method](https://github.com/alexreisner/geocoder/blob/f905e455273ae1489e7af3c006caf4ecd7cd6fe7/lib/geocoder/request.rb#L71-L73) expects IPv4 addresses. When passed an IPv6 address, it incorrectly tries to remove the port by splitting on `:`. The syntax for specifying a port with an IPv6 address is specified in [RFC 3986](http://www.ietf.org/rfc/rfc3986.txt). An example:

    [3ffe:1900:4545:3:200:f8ff:fe21:67cf]:80

abevoelker:
Closes #1210

### DIFF
--- a/lib/geocoder/request.rb
+++ b/lib/geocoder/request.rb
@@ -81,7 +81,18 @@ module Geocoder
     end
 
     def geocoder_remove_port_from_addresses(ip_addresses)
-      ip_addresses.map { |ip| ip.split(':').first }
+      ip_addresses.map do |ip|
+        # IPv4
+        if ip.count('.') > 0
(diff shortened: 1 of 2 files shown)


alexreisner:
Thanks for this!

abevoelker:
No problem! :bowing_man:

PR #1254, opened 2018-01-10
Provide ipdata.co as a lookup

roschaefer:
Hello everybody!

Here's another ip->geolocation endpoint. Most of it is inspired ~copy&pasted~ by the freegeoip example.

Side notes:
1) I see a lot of deprecation warnings when in run `rake test`
2) There is a failing spec on the latest commit of the master branch: 
```
===============================================================================================================================================================================================
/home/robert/Development/Ruby/geocoder/test/unit/lookups/yandex_test.rb:29:in `test_yandex_query_url_contains_bbox'
     26:       "Some Intersection",
     27:       :bounds => [[40.0, -120.0], [39.0, -121.0]]
     28:     ))
  => 29:     assert_match(/bbox=40.0+%2C-120.0+%7E39.0+%2C-121.0+/, url)
     30:   end
     31: end
Failure: test_yandex_query_url_contains_bbox(YandexTest):
  </bbox=40.0+%2C-120.0+%7E39.0+%2C-121.0+/> was expected to be =~
  <"https://geocode-maps.yandex.ru/1.x/?bbox=40.000000%2C-120.000000~39.000000%2C-121.000000&format=json&geocode=Some+Intersection&key=%5B%22auti-id%22%2C+%22auth-token%22%5D&plng=en">.
==============================================================================================================================================================================================
```
when I replace `%7E` with `~` the test runs fine. I guess, it's because the API has changed since the last time the master was built?

### DIFF
--- a/README.md
+++ b/README.md
@@ -856,6 +856,18 @@ This uses the PostcodeAnywhere UK Geocode service, this will geocode any string 
 * **Documentation**: https://db-ip.com/api/doc.php
 * **Terms of Service**: https://db-ip.com/tos.php
 
+#### Ipdata.co (`:ipdata_co`)
+
+* **API key**: optional, see: https://ipdata.co/pricing.html
+* **Quota**: 1500/day (up to 600k with paid API keys)
+* **Region**: world
+* **SSL support**: yes
+* **Languages**: English
+* **Documentation**: https://ipdata.co/docs.html
+* **Terms of Service**: https://ipdata.co/terms.html
+* **Limitations**: ?
+
+
 ### IP Address Local Database Services
 
 #### MaxMind Local (`:maxmind_local`) - EXPERIMENTAL
--- a/lib/geocoder/lookup.rb
+++ b/lib/geocoder/lookup.rb
@@ -71,6 +71,7 @@ module Geocoder
         :maxmind_geoip2,
         :ipinfo_io,
         :ipapi_com,
+        :ipdata_co,
         :db_ip_com
       ]
     end
--- a/lib/geocoder/lookups/ipdata_co.rb
+++ b/lib/geocoder/lookups/ipdata_co.rb
@@ -43,5 +43,14 @@ module Geocoder::Lookup
     def host
       "api.ipdata.co"
     end
+
+    def check_response_for_errors!(response)
+      if response.code.to_i == 403
+        raise_error(Geocoder::RequestDenied) ||
+          Geocoder.log(:warn, "Geocoding API error: 403 API key does not exist")
+      else
+        super(response)
+      end
+    end
   end
 end
--- a/lib/geocoder/lookups/ipdata_co.rb
+++ b/lib/geocoder/lookups/ipdata_co.rb
@@ -19,7 +19,7 @@ module Geocoder::Lookup
     private # ---------------------------------------------------------------
 
     def results(query)
-      Geocoder.configure(:http_headers => { "api-key" => configuration.api_key }) if configuration.api_key
+      Geocoder.configure(:ipdata_co => {:http_headers => { "api-key" => configuration.api_key }}) if configuration.api_key
       # don't look up a loopback address, just return the stored result
       return [reserved_result(query.text)] if query.loopback_ip_address?
       # note: Ipdata.co returns plain text on bad request
--- a/lib/geocoder/lookups/ipdata_co.rb
+++ b/lib/geocoder/lookups/ipdata_co.rb
@@ -1,0 +1,51 @@
+require 'geocoder/lookups/base'
+require 'geocoder/results/ipdata_co'
+
+module Geocoder::Lookup
+  class IpdataCo < Base
+
+    def name
+      "ipdata.co"
+    end
+
+    def supported_protocols
+      [:https]
+    end
+
+    def query_url(query)
+      "#{protocol}://#{host}/#{query.sanitized_text}"
+    end
+
+    private # ---------------------------------------------------------------
+
+    def parse_raw_data(raw_data)
+      raw_data.match(/^<html><title>404/) ? nil : super(raw_data)
+    end
+
+    def results(query)
+      # don't look up a loopback address, just return the stored result
+      return [reserved_result(query.text)] if query.loopback_ip_address?
+      # note: Ipdata.co returns plain text on bad request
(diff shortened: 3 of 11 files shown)


roschaefer:
The build server fails on jruby because of:
```
3.51s$ bundle install --jobs=3 --retry=3
The program 'bundle' is currently not installed. To run 'bundle' please ask your administrator to install the package 'bundler'
The command "eval bundle install --jobs=3 --retry=3 " failed. Retrying, 2 of 3.
```
I guess that's unrelated?

alexreisner:
This looks great! Thanks. (And yeah, those JRuby failures are unrelated.) The only issue I see is error handling. For example, if I search for an invalid IP address I get a string which your current `parse_raw_data` will not catch:

    curl https://api.ipdata.co/8.8.8.
    8.8.8. does not appear to be an IPv4 or IPv6 address

I think `parse_raw_data` should handle that one too.

alexreisner:
Excellent! I love it when a solution *only removes* code. However, are there other errors we should handle differently? Seems like it might be helpful to return a specific message on, say, authentication failure.

roschaefer:
@alexreisner thanks for pointing out. It's quite silly to check for HTML tags if the server responds only JSON. I believe the check in the original freegeoip lookup is pointless, too.

Just out of curiosity I tried IPv6 addresses:
```
➜  geocoder git:(master) curl https://api.ipdata.co/be37:44bd:8630:7a0:2a3d:ff95:dd33:42f0
{"message": "Internal server error"}%   
```
 Just contacted the maintainer of ipdata.co.

I don't know what *should* happen as I don't know what information can be derived from IPv6 addresses. At least, the server shouldn't crash.

roschaefer:
I agree that a specific message on authentication failure would be great. But that would probably require some refactoring in the base lookup class.

I also found it quite suprising to force ip lookup with the options hash. Otherwise you get a "google" result object. :confused:

This is a different PR maybe?

alexreisner:
Why would it require refactoring of the base lookup? I was assuming it would be as easy as checking a status code but maybe not. Looking at [the docs](https://ipdata.co/docs.html), it looks like they only return 200, 400 (invalid input), and 429 (over quota). Maybe we should just handle those?

Also, how would an API key work in your implementation? Unfortunately I can't see from Ipinfo.co's docs how it's supposed to work...

roschaefer:
@alexreisner sorry, not the base lookup class, but `test/test_helper.rb`. I got confused, as test helper is monkey patching methods of the base lookup class.

Regarding API keys:
According to the python library, all you have to do is to set one HTTP request header:
https://github.com/ipdata/python/blob/master/ipdata/ipdata.py
e.g.
```
curl https://api.ipdata.co/141.3.135.0 --header "api-key: XXXX" -I
```
A wrong API key will return 403 status code.


I'm having a hard time to debug the custom HTTP mocking framework used in this repository. In other repositories, I'm using [vcr](https://github.com/vcr/vcr). It's convenient, reliable and does not require monkey patching. It allows you to capture the request and store the response on disk.

Is there any way to make and actual request and capture it? I've seen, s.b is already using WebMock for google, but in that case it's testing a private method. How can I write a test case for an *optional* api key? As far as I can tell, the other endpoints have the api key in the URL, ist that right? How can I set an HTTP header? The code to check for HTTP status codes is located in the base lookup class [here](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookups/base.rb#L250). It matches the same status codes responses of ipdata.co (400, 429 ..). How can I create a test case for that?

alexreisner:
Hmm...good points/questions. Geocoder's ad hoc mocking framework is not good. I'm open to using something else, but there are at least a couple of issues:

1. The need to hide credentials. Eg: is there a way with VCR to hide sensitive parts of an HTTP request (eg, API key) and still match/send the request as desired? When I looked into it a couple of years ago this didn't seem possible.
2. There are certain API responses that we need to test which only occur under certain conditions, like [when a quota or rate limit is exceeded](https://github.com/alexreisner/geocoder/blob/master/test/fixtures/location_iq_over_limit). It's unclear how to get those responses when sending real HTTP requests (eg, with VCR).

If there are solutions to those issues then I'm open to considering VCR again. As for your questions:

1. You can configure `:http_headers` on a per-lookup basis (like everything else in the global config), and while it's a little ugly, merging `api-key` into the existing hash seems like an easy way to do this without rewriting the base lookup. (Does that make sense?)
2. I totally forgot about that universal error-checking in the base lookup. Nothing to do there! To test you can use something like `MockHttpResponse.new(body: "...", code: "429")`, but adding this test is not necessary for you to do here (we should really have this tested for the base lookup).

roschaefer:
@alexreisner regarding your questions:

1. You just edit the stored cassette and replace the real API key with 'XXX'
2. True, but you would face the same problem without vcr. You could use webmock (which is used by vcr internally) and ask the support of @ipdata for a sample response.

I configure the `:http_header` in `results()` which seems to be the entry point of other lookup classes for additional checking before sending out the request. I have to admit, it looks ugly and I feel bad about it.

alexreisner:
Thanks for this. However, a couple of things before I merge:

1. When calling `Geocoder.configure(:http_headers, ...)` we don't want to set the headers at the root level, because it will affect any other configured lookups. I think it should be `Geocoder.configure(:ipdata_co => {:http_headers => ...})`.
2. Really appreciate your work on `test_api_key`, but until we come up with a better global solution, let's leave this untested (with maybe a TODO in the code). I'd rather not add dependencies or have tests making network calls before evaluating how WebMock can be used for testing all lookups. Does that make sense?

roschaefer:
Long time no see.

Removed test case as @alexreisner suggested. We have tests using WebMock [here](https://github.com/alexreisner/geocoder/blob/master/test/unit/lookups/ban_data_gouv_fr_test.rb#L115) and [here](https://github.com/alexreisner/geocoder/blob/master/test/unit/lookups/google_test.rb#L118), though.

jonathan-kosgei:
Hey @alexreisner  any feedback on this?

alexreisner:
Merged! Sorry for the delay. Thanks very much @roschaefer for the good work on this.

alexreisner:
Thanks @jonathan-kosgei. Updated: 8c919117370284c5f65a7d5e4c272f69d6c9790e

PR #1255, opened 2018-01-11
Fix Travis CI build errors. Closes #1224

### LINKED ISSUE #1224: Travis Tests failing
Travis tests are failing for some of the builds in matrix due to gem dependency issues causing bundle install not to complete correctly See: https://travis-ci.org/alexreisner/geocoder/builds/277296105

abevoelker:
Closes #1224. There were a few main causes:

* bundler was missing on jruby-19mode in Travis CI's image so I slightly modified a suggested fix from here: https://github.com/travis-ci/travis-ci/issues/5578#issuecomment-180445244
* the `pg` gem released a new 1.0.0 version yesterday which caused issues with [Rails versions < 5 requiring `~> 0.11` at runtime](https://github.com/rails/rails/blob/0cad778c2605a5204a05a9f1dbd3344e39f248d8/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L13), and [Rails 5 requiring `~> 0.18`](https://github.com/rails/rails/blob/71f4758c50fca5220429365c23b014458e7341f6/activerecord/lib/active_record/connection_adapters/postgresql_adapter.rb#L2) (figuring this out was slightly complicated by ruby/rake#247)
* the `activerecord-jdbc-adapter` gem released version 50.0 semi-recently, which should only be used by Rails 5+

Also, I don't know whether the intention is for the Gemfile in the root of the project to run agains the latest version of Rails or not, but `mongoid` being pinned to 2.6.0 is forcing it to select Rails 3, which is why I had to apply the version restriction on `activerecord-jdbcpostgresql-adapter` there to get the tests to pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,8 @@ gemfile:
   - gemfiles/Gemfile.rails3.2
   - gemfiles/Gemfile.rails4.1
   - gemfiles/Gemfile.rails5.0
+before_install:
+  - which bundle >/dev/null 2>&1 || gem install bundler
 matrix:
   exclude:
     - rvm: 1.9.3
--- a/Gemfile
+++ b/Gemfile
@@ -31,14 +31,14 @@ group :test do
   gem 'webmock'
 
   platforms :ruby do
-    gem 'pg'
+    gem 'pg', '~> 0.11'
     gem 'mysql2', '~> 0.3.11'
   end
 
   platforms :jruby do
     gem 'jdbc-mysql'
     gem 'jdbc-sqlite3'
-    gem 'activerecord-jdbcpostgresql-adapter'
+    gem 'activerecord-jdbcpostgresql-adapter', '~> 1.3.0'
   end
 end
 
--- a/Rakefile
+++ b/Rakefile
@@ -32,7 +32,7 @@ namespace :db do
 
     desc 'Drop the MySQL test databases'
     task :drop do
-      `mysqladmin --user=#{config['mysql']['username']} -f drop #{config['mysql']['database']}`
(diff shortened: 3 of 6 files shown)


alexreisner:
Thanks for this! Really glad to have the tests fixed. Also you bring up a good point about Mongoid being pinned to 2.6.0. That was for Ruby 1.8 compatibility, which has since been dropped. So we don't need to pin it anymore. You want to fix that too?

abevoelker:
Sure; I tested that change in a separate branch, however that surfaced [one new build error](https://travis-ci.org/abevoelker/geocoder/jobs/327730693) which will require a new version of the `activerecord-jdbcpostgresql-adapter` to be published: https://github.com/jruby/activerecord-jdbc-adapter/issues/780#issuecomment-356982735

The build error could probably be worked around right now by changing the Gemfile to use the GitHub source instead:

```
gem 'activerecord-jdbcpostgresql-adapter', github: 'jruby/activerecord-jdbc-adapter'
```

But rather'n dirtying up the Gemfile it's probably preferable to just wait for the new gem version to be cut.

alexreisner:
What about locking `mongoid` and `activerecord-jdbcpostgresql-adapter` so the most recent combo that works? I misspoke when I said we don't need to pin it at all--if we don't, the build constantly breaks (as illustrated). (Sorry!)

abevoelker:
No preference here, it's your show :smile: 

Unfortunately I can't put more time on this PR; I'll have to bow out now as I've ran out of personal time. I leave it in your (or someone else's) capable hands to tidy up the Gemfile to your liking :v:

alexreisner:
Sounds good--I'll finish this up. Appreciate the help!

PR #1256, opened 2018-01-15
Added :host to list of allowed config params

manzhikov:


alexreisner:
What problem does this solve?

manzhikov:
@alexreisner this allow to setup `:host` for _Nominatim_

alexreisner:
Ah, yes. But you should be able to add `:host` to the `:nominatim` config block without any change to the root-level config. `:host` only applies to a few lookups so I don't want to make it a global setting.

Issue #841, opened 2015-04-16
Bing lookup doesn't raise an exception for some kinds of server errors


jamesgecko:
If Bing's API fails to return something expected, the Bing lookup [just prints a warning](https://github.com/alexreisner/geocoder/blob/90e2ef1c82685c3a16bffb2c42e97bc210922f1b/lib/geocoder/lookups/bing.rb#L48). An exception would be more useful, especially if the server is complaining about a rate limit.

I glanced through the other lookup classes, and the [MapQuest](https://github.com/alexreisner/geocoder/blob/7143136f35bab70966abcb22c23933bce79c970b/lib/geocoder/lookups/mapquest.rb#L52) and [Google](https://github.com/alexreisner/geocoder/blob/90e2ef1c82685c3a16bffb2c42e97bc210922f1b/lib/geocoder/lookups/google.rb#L38) ones do raise exceptions.

alexreisner:
Sounds great. Would love a pull request for raising exceptions on various error conditions with Bing.

peteb:
Might this be related to #840? I'm using Bing and it seems to be caching nil.

alexreisner:
@peteb it looks related in that Bing doesn't use HTTP response codes well (which is part of the cause of  both issues), but a solution to this won't fix #840 (or vice versa).

germsvel:
@alexreisner how would you feel about raising a `Geocoder::Error` for a 403, 500, and 501? 

Bing seems to [support those](https://msdn.microsoft.com/en-us/library/ff701703.aspx), and at least it would notify the users of the gem that something is wrong. If you feel like that would be an improvement over just printing the warning, I'd be happy to take a stab at it and open a PR.

alexreisner:
I'm still open to a PR for this, but note that each of those response codes implies a different exception. Maybe something like this:

403: `Geocoder::RequestDenied`
500: `Geocoder::ServiceUnavailable`
503: `Geocoder::ServiceUnavailable`

but it really depends on the conditions under which Bing returns those codes (I'm going by generic HTTP definitions). Also, please be sure to use the `raise_error` method (see [the base lookup](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookups/base.rb) for an example).

germsvel:
Thanks for the response @alexreisner. And thanks for mentioning `raise_error`. I will certainly use that. 

[...]

PR #1259, opened 2018-01-17
Raise bing errors for statuses 403, 500, 503

germsvel:
This attempts to fix https://github.com/alexreisner/geocoder/issues/841. 

Summary
=======

Bing returns the status codes within the json response. When a request is forbidden, it will return a 403. If there is an internal server error, it will return a 500. And if the service is unavailable for some other reason, it will return a 503. That information was obtained from bing's [Status Codes and Error Handling](https://msdn.microsoft.com/en-us/library/ff701703.aspx) documentation.

In this commit, we handle those three status codes by raising `Geocoder::RequestDenied` (for 403) and `Geocoder::ServiceUnavailable` (for 500 and 503). This allows for the users of the Geocoder to handle those errors in their applications.

## Note on fixtures

I added two fixtures to the tests, one that returns a 403 and another that returns a 500. These two were copied from other bing fixtures and are not real error responses from bing. If those need to be real responses obtained in the wild, I can look more into how to get those. Unfortunately, no sample responses from bing's [Location's Api documentation](https://msdn.microsoft.com/en-us/library/ff701711.aspx) include the cases for errors.


alexreisner:
Excellent pull request! Thanks!

Issue #1247, opened 2017-12-03
Does not work with Rails 5.2


morgoth:
### Expected behavior

MyClass.near works as before

### Actual behavior

MyClass.near raises
```
     NoMethodError:
       undefined method `near_scope_options' for #<ActiveRecord::Relation []>
```

It's happening because https://github.com/alexreisner/geocoder/blob/5353fea4c963d4db2f429f5c343a202f28109266/lib/geocoder/stores/active_record.rb#L119 is a private method.

I workaround it by 
```ruby
class << self
  public :near_scope_options
end
```

### Environment info

* Geocoder version: 1.4.5
* Rails version: 5.2.0.beta2

stevenharman:
What in Rails changed that would trigger this? `Geocoder::Store::ActiveRecord::near_scope_options` has been private for years. @alexreisner Do you have a recommended workaround until a fix can be release?

mbaird:
Was caused by https://github.com/rails/rails/pull/29301. Also see https://github.com/rails/rails/issues/31740.

stevenharman:
Sounds like one of use should open a PR to make `near_scope_options` public then? Or would @alexreisner rather make the change? We'll need a new release cut, either way.

alexreisner:
I'm very much open to a PR, if anyone wants to tackle this!

PR #1269, opened 2018-02-27
Make #near_scope_options public for Rails 5.2

stevenharman:
Rails 5.2 changed the receiver within a scope from `klass` to `relation`. With that, we no longer have private access to `Geocoder::Store::ActiveRecord#near_scope_options`, and so must make it public.

see: https://github.com/rails/rails/pull/29301

Closes: #1247


alexreisner:
Thanks for this!

PR #1272, opened 2018-03-08
Explain that 20 is default radius for near()

kriskhaira:
20 was used as an example radius value in the README for near(), but it isn't explicitly mentioned that it's the default. I had to look through the code to find out, and found it here:
https://github.com/alexreisner/geocoder/blob/bab2e367880b02505dea077951bfada987e2c93d/lib/geocoder/stores/active_record.rb#L117

It would be nice if the README was clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,10 +156,10 @@ Please use MongoDB's [geospatial query language](https://docs.mongodb.org/manual
 
 To find objects by location, use the following scopes:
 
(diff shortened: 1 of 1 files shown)


alexreisner:
Thanks for this. I agree. Will merge, but please fix mismatched "40"s (in comment) and "50"s (in code).

kriskhaira:
@alexreisner Thanks for pointing it out. I've fixed it.

Issue #1273, opened 2018-03-08
Custom params don't work?


Meekohi:
### Expected behavior
Using Geocodio, and would like to pass `&fields=timezone` in the query.

First I tried:
`geocoded_by :address, :params => {:fields => "timezone"}` in the model but

Also tried:
`  :params       => {"fields" => "timezone"},` in `geocoder.rb`

both of these did not include those params in the request.

### Actual behavior

Should add `&fields=timezone` in the query to query url.

### Steps to reproduce

see above

### Environment info

* Geocoder version: 1.4.6
* Rails version: 4.2.8
* Database (if applicable): n/a

Meekohi:
Pretty stumped. It doesn't seem like passing in params via the `configuration` works at all. Printing out the whole `ConfigurationHash` object doesn't show that `params` is included (triggers `method_missing` and crashes)

PR #1276, opened 2018-03-19
Update Geocodio to v1.3

Meekohi:
Update to Geocodio API v1.3 (https://geocod.io/docs/#v1-3-released-on-march-12th-2018).
Always request additional `timezone` information (open to other ways to accomplish this, but see https://github.com/alexreisner/geocoder/issues/1273).


alexreisner:
Thanks for this. The custom params should be fixed by d94b987. I'm open to updating Geocodio to 1.3 but that should be in its own pull request.

As an aside, it's worth noting that 1.3 includes breaking changes, as did 1.1 and 1.2. This seems to be a new habit of Geocodio's. We might want to consider adding a way to specify API version number so people aren't locked into a version they might not want to use.

PR #1282, opened 2018-04-05
Use new /location endpoint for Telize

dmchoull:
Telize is moving from their current "/geoip" endpoint to a new "/location" endpoint. [See the announcement here.](https://market.mashape.com/fcambus/telize/announcements/2)

The /geoip endpoint will stop receiving database updates as of May 1st, and will stop responding to queries as of June 1st, so it is important to get this updated quickly for all Telize users.

The new endpoint does return data in a slightly different format, as detailed here: https://www.telize.com/telize-2.0.0-has-been-released/


alexreisner:
Thanks for this!

PR #1283, opened 2018-04-06
Add the possibility to set a method as a query option

the7oka:
I had a problem with validating addresses because sometime random numbers will actually find an address some where.
My initial idea was to narrow the scope of the search to the country. but that was impossible to change with each search since the params are loaded on the server startup.

Options should be dynamic and can be changed with each search request.

### DIFF
--- a/lib/geocoder/stores/base.rb
+++ b/lib/geocoder/stores/base.rb
@@ -92,7 +92,11 @@ module Geocoder
 
         query_options = [:lookup, :ip_lookup, :language, :params].inject({}) do |hash, key|
           if options.has_key?(key)
-            val = options[key]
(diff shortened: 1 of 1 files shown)


alexreisner:
Options *are* dynamic and can be changed on each request. Unless I'm misunderstanding what you're asking...

(Sorry for the slow response.)

alexreisner:
I see what you're suggesting, but it's already possible to use an anonymous method (lambda) to do this, and I'm worried (with your solution) about the potential collision of lookup names and the object's method names. As shown in [the README](https://github.com/alexreisner/geocoder#advanced-model-configuration) you can call a method by name via lambda function.

(I've been working on clarifying the README lately, since a lot of stuff like this has been buried in it.)

PR #1289, opened 2018-05-01
Deprecate Freegeoip and Implement Ipstack

### LINKED ISSUE #1279: Freegeoip Deprecations
First off, thanks for this great project!!

Secondly, has there been any work done to handle the **Important Announcement** section which was outlined on the freegeoip github page [https://github.com/apilayer/freegeoip/](https://github.com/apilayer/freegeoip/)?  Most importantly that an api key will now be required, to use the service.

Since freegeoip is the default IP lookup strategy, I just wanted to see if this was on the radar to be upgraded, or if I just missed a solution that allows to me to add my api_key.

Heath101:
Fixes #1279 

One of the unique problems here is that regardless of whether people update versions of this gem, the functionality for freegeoip will just stop.  We can introduce as many mechanisms in this PR to help with that, but there will probably be quite a few people for whom they wake up July 1st and their project just doesn't work.

### Post Install Message & Wiki
The post install message contains a link to a wiki page (currently just empty) on how to upgrade from freegeoip to Ipstack. I have a draft of what the wiki page could contain on my forks wiki section [here](https://github.com/Heath101/geocoder/wiki/Freegeoip-Discontinuation-Upgrade-Instructions).  I would love as a part of reviewing this PR input on the content of that wiki page as well.  Then when it is ready we can replace the dummy one that I already created [here](https://github.com/alexreisner/geocoder/wiki/Freegeoip-Discontinuation-Upgrade-Instructions)

### Features Not Implemented
There are still a few features of the ipstack api that I have not implemented in this PR.  **Batching**, **JSONP support**, and **Requester IP lookup**.  Since these were not available with freegeoip, I kinda just wanted to get in the stuff that I thought we be the most useful right off the bat.

### Notes on this PR
I noticed that the initializer template still had a reference to freegeoip as the ip_lookup, so I changed that to the new default `:ipinfo_io`.

I tried to keep the object api of the freegeoip and ipstack result objects the same.  There is only one difference and that is that ipstack does not provide a `metro_code`.  This change is noted in the wiki upgrade page.

I did make a slight deviation from one of the patterns I saw in the current code base.  The current code base allows you to pass in params as options to a lookup.  I decided that I wanted to hide the actual values of the params, in favor of a small DSL.  For example:

```ruby
# not implemented this way
Geocoder.search('0.0.0.0', {params: {security: '1'} })

# instead using a small dsl
Geocoder.search('0.0.0.0', {security: true })
```

This DSL is described in the README section for the Ipstack lookup.  If this deviates too much from the params pattern, then I can revisit that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -763,15 +763,17 @@ This uses the PostcodeAnywhere UK Geocode service, this will geocode any string 
 * **Documentation**: https://ipstack.com/documentation
 * **Terms of Service**: ?
 * **Limitations**: ?
-* **Notes**: To use Ipstack set `Geocoder.configure(:ip_lookup => :ipstack, :api_key => "your_ipstack_api_key")`. Support for the ipstack JSONP, batch lookup, and requester lookup features has not been implemented yet.  DSL options are as follows:
+* **Notes**: To use Ipstack set `Geocoder.configure(:ip_lookup => :ipstack, :api_key => "your_ipstack_api_key")`. Support for the ipstack JSONP, batch lookup, and requester lookup features has not been implemented yet.  This lookup does support the params shown below:
 * ```ruby
     # Search Options Examples
 
-    Geocoder.search('0.0.0.0', {
-        hostname: true, # enable hostname in response
-        security: true,
-        fields: ['fields','go','here'],
-        language: 'en',
+    Geocoder.search('0.0.0.0', { 
+        params: {
+            hostname: '1', 
+            security: '1', 
+            fields: 'country_code,location.capital,...', # see ipstack documentation
+            language: 'en'
+        }
     })
     ```
 
--- a/README.md
+++ b/README.md
@@ -753,7 +753,29 @@ This uses the PostcodeAnywhere UK Geocode service, this will geocode any string 
 
 ### IP Address Services
 
-#### FreeGeoIP (`:freegeoip`)
+#### Ipstack (`:ipstack`)
+
+* **API key**: required - see https://ipstack.com/product
+* **Quota**: 10,000 requests per month (with free API Key, 50,000/day and up for paid plans)
+* **Region**: world
+* **SSL support**: yes ( only with paid plan )
+* **Languages**: English, German, Spanish, French, Japanese, Portugues (Brazil), Russian, Chinese
+* **Documentation**: https://ipstack.com/documentation
+* **Terms of Service**: ?
+* **Limitations**: ?
+* **Notes**: To use Ipstack set `Geocoder.configure(:ip_lookup => :ipstack, :api_key => "your_ipstack_api_key")`. Support for the ipstack JSONP, batch lookup, and requester lookup features has not been implemented yet.  DSL options are as follows:
+* ```ruby
+    # Search Options Examples
+
+    Geocoder.search('0.0.0.0', {
+        hostname: true, # enable hostname in response
+        security: true,
+        fields: ['fields','go','here'],
+        language: 'en',
+    })
+    ```
+
+#### FreeGeoIP (`:freegeoip`) - DEPRECATED ( see [ upgrade instructions](https://github.com/alexreisner/geocoder/wiki/Freegeoip-Discontinuation-Upgrade-Instructions) )
 
 * **API key**: none
 * **Quota**: 15,000 requests per hour. After reaching the hourly quota, all of your requests will result in HTTP 403 (Forbidden) until it clears up on the next roll over.
--- a/README.md
+++ b/README.md
@@ -777,7 +777,7 @@ This uses the PostcodeAnywhere UK Geocode service, this will geocode any string 
     })
     ```
 
-#### FreeGeoIP (`:freegeoip`) - DEPRECATED ( see [ upgrade instructions](https://github.com/alexreisner/geocoder/wiki/Freegeoip-Discontinuation-Upgrade-Instructions) )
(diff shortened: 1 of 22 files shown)


alexreisner:
Thanks @Heath101, this is great! I especially appreciate all the error handling (and testing). A couple of comments:

1. The wiki page is a good idea, however I'm not really keen on recommending Ipstack over any other provider. Its only relation to Freegeoip is business. I think the more important transition to document is actually to (the new default) IpInfo. So in my opinion that's what should be on the wiki page. And the post-install message probably shouldn't refer to "upgrading" as much as "switching."
2. I like your thinking on the DSL, but I've rejected this kind of thing in the past because I felt (and still do) that it's more important to maintain consistency among lookups so that we can switch from one to the other with as few surprises as possible.

Aside from that, this looks awfully close to merge-able. Nice work!

Heath101:
@alexreisner Ok, I can remove the DSL and follow the params paradigm no problem.  I also totally didn't notice the `coordinates` method on the base result class, so I'll update the ipstack result object to follow that "interface" as well ( and remove warning logging on the other result methods ).

I'll also change the verbiage to switching.  I think the main reason I referred to it that way was because, at least for my use case, we could not use `:ipinfo_io` because of the non-commercial requirement.  So the default lookup's requirements are becoming a little tighter.  I'd like to at least mention the non commercial stipulation in the wiki, only to prevent those who move to the new default from implicitly agreeing to something that may violate the use of the ipinfo.io api.

I'll add another comment after I've updated the wiki, also I plan on changing the name of the wiki page to be more generic.

alexreisner:
That all sounds good! That's an excellent point about Ipinfo's non-commercial requirement. *Definitely* worth mentioning that on the wiki, though I'm still hesitant to recommend Ipstack. (To expand on that, in case not obvious: I want to avoid endorsing any particular API. Defaults are chosen based solely on what will provide the best out-of-box experience to the new user. Ultimately everyone should learn about all the options and choose what's most appropriate, though I know in reality a lot of people won't do that. So yeah, let's definitely mention the limitations of Ipinfo.)

Thanks again for this!

Heath101:
@alexreisner I have made changes to the wiki as well.  Found here

[https://github.com/Heath101/geocoder/wiki/Freegeoip-Discontinuation](https://github.com/Heath101/geocoder/wiki/Freegeoip-Discontinuation)

demitriousj:
Just inquiring, is there anything in particular holding up this PR?

alexreisner:
@demitriousj nothing particular--just reviewed and merged it. Going to shoot for a new gem release in the coming week.

@Heath101 thanks for the good work!!

MaicolBen:
@Heath101 why did you reused the `api_key` of the `lookup`? Now you can't set api key for lookup (Google in my case) and ip_lookup at the same time. 

@alexreisner I see this behavior for other ip_lookup as well, can't we define a `ip_lookup_api_key` instead?

alexreisner:
@MaicolBen sorry this is unclear. You can set any config variable for an individual lookup by using that lookup's name as a hash key, eg:

    Geocoder.configure(
      lookup: :google,
      ip_lookup: :ipstack,
      google: {
        api_key: "..."
      },
      ipstack: {
        api_key: "..."
      }
    }

MaicolBen:
I didn't know, thank you very much!

tmferretti:
Hey @alexreisner, I have followed the correct structure for the configuration: 
`
Geocoder.configure(
  timeout: 8,                 # geocoding service timeout (secs)
  lookup: :google,            # name of geocoding service (symbol)
  ip_lookup: :ipinfo_io,      # name of IP address geocoding service (symbol)
  ipinfo_io: {
    api_key: '[APIKEY]'
  },
  use_https: true,           # use HTTPS for lookup requests? (if supported)
)
`
My company upgraded our account on inpinfo.io and I have implemented the correct api key that is associated with that account. However, no hits are registering from the ipinfo account dashboard while using geocoder's ip lookup, but they do register when i manually call to the endpoint with said key.

alexreisner:
@tmferretti Are you getting any results? Can you copy/paste the results of a console `Geocoder.search` command here?

tmferretti:
@alexreisner I am:
```
2.2.7 :001 > Geocoder.search("50.78.167.161")
 => [#<Geocoder::Result::IpinfoIo:0x007fafb6b01540 @data={"ip"=>"50.78.167.161", "city"=>"", "region"=>"", "country"=>"US", "loc"=>"37.7510,-97.8220"}, @cache_hit=nil>]

```
 I believe its going through ipinfo_io's free verision.

alexreisner:
Let's look at the URL Geocoder is using. Can you post the output of this?:

    Geocoder::Lookup.get(:ipinfo_io).query_url(Geocoder::Query.new("50.78.167.161"))

tmferretti:
```
2.2.7 :001 > Geocoder::Lookup.get(:ipinfo_io).query_url(Geocoder::Query.new("50.78.167.161"))
 => "https://ipinfo.io/50.78.167.161/geo?token=[API_KEY]"
```

alexreisner:
That looks right to me, according to Ipinfo's [docs](https://ipinfo.io/developers). What URL were you using when you were calling the API manually?

alexreisner:
Also, I assume you're replacing `[API_KEY]` with your actual key?

tmferretti:
@alexreisner, Yeah I was. We use a private repo so it is that string and not an environment variable. Apologies for the late response, but the key was for ipstack and not ipinfo_io.. womp womp

Issue #1291, opened 2018-05-10
Make maxmind_geoip2 result return time zone


yogodoshi:
Hello, I'm not very familiarized with contributing to open source so I'm sorry if I'm breaking any protocols or not following best practices 😛

That being said, I wanted to get the time_zone from `Geocoder::Result::MaxmindGeoip2`. From [their docs](https://dev.maxmind.com/geoip/geoip2/web-services/):
<img width="908" alt="screen shot 2018-05-11 at 2 20 38 am" src="https://user-images.githubusercontent.com/412144/39889414-03011f40-54c2-11e8-91ad-1ca853c1ca96.png">

Are you planning on implementing it really soon or should/can I open a PR for it? :)

And thanks for making this awesome gem 👍

alexreisner:
Thanks for this! I would accept a PR which adds a `time_zone` method. But note that even without a specific method you can access the data using: `result.data['time_zone']`.

yogodoshi:
@alexreisner I tried but the data method is private, would you prefer a PR to make it public? 🤔

alexreisner:
Oh, I see. That's a bug, and yes, a PR would be great. All that needs to be done is to delete that method from the result class. It's already defined [in the base class](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/results/base.rb#L6).

SpiffyStores:
This seems to cause a bug when using maxminddb. I haven't had time to investigate fully, but the MaxMindDB::Result object does not support fetch, so the removal of the to_hash breaks things when called from geocoder.

alexreisner:
@SpiffyStores you're right, thanks. In f8d5eeb247c42d71dc7d5f9191d47484c2e469be I've restored `#data` as a public method.

PR #1292, opened 2018-05-11
Remove Geocoder::Results::Geoip2#data

yogodoshi:
As requested and discussed on #1291, removing `data` method so we can access the `@data` variable made available on `Geocoder::Result::Base`.

Ran the specs locally and none started failing after the change (had one already failing before the change).


alexreisner:
Thanks! (And yes, the build error is unrelated.)

PR #1296, opened 2018-05-31
Sensis support

zhaomengru2015:
implementation for [sensis support](https://github.com/alexreisner/geocoder/issues/603)


alexreisner:
@zhaomengru2015 Thanks for this! And welcome to the project. However, I have some bad news. I don't think Sensis is a good candidate for addition to Geocoder. To justify making changes to the base lookup I think we need a service that provides more than address geocoding in one country. Does Sensis have much better coverage of Australia than other solutions? Or other features that make it indispensable in AU?

DanielHeath:
Sensis is run by the now-private (formerly government-monopoly) telco in Australia, so it has very accurate geodata to draw on.

PR #1301, opened 2018-06-15
Add config disclaimer and link

jpowell:
Readme is long and geocoding service configuration instructions are far down the document. Thought this would help.

A different solution could be adding a table of contents, and / or breaking up the readme into wiki files. This readme is just so long.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,7 @@ and run at the command prompt:
 
     bundle install
 
+**You must [configure your geocoding service](#geocoding-service-lookup-configuration) before using**
 
 Object Geocoding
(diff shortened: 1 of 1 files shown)


alexreisner:
Thanks for this! You're right about the README being long, and the geocoding services way down there. However, I think a table of contents is a better solution. It's important that the gem "works" (in a basic way) for new users, without any configuration, so I don't want to suggest that people need to read the most technical section of the README before using the gem.

jpowell:
agreed

alexreisner:
Note 1d5ad37

PR #1305, opened 2018-06-26
Add error details

JonRowe:
When Google returns a geocoding error, there is often helpful debugging information in the error message which is currently silently swallowed, this adds that text to the error handling.

Use case: Our API key had referrer enforcement turned on, suddenly today Google cut off out server side API as not allowed, this was unexpected and all we got from the server was request denied, manually curling the api cause the correct details to be exposed:

```
{
   "error_message" : "API keys with referer restrictions cannot be used with this API.",
   "results" : [],
   "status" : "REQUEST_DENIED"
}
```


alexreisner:
Thanks!

Issue #1258, opened 2018-01-17
Stubbed results do not maintain consistency between coordinates and lat, long


derekprior:
The README contains the following code for stubbing Geocoder results:

```ruby
Geocoder.configure(:lookup => :test)

Geocoder::Lookup::Test.add_stub(
  "New York, NY", [
    {
      'coordinates'  => [40.7143528, -74.0059731],
      'address'      => 'New York, NY, USA',
      'state'        => 'New York',
      'state_code'   => 'NY',
      'country'      => 'United States',
      'country_code' => 'US'
    }
  ]
)
```

### Expected behavior

When using the stub above for a test that attempts to Geocode an ActiveRecord model via the `geocode` method should succeed and set `latitude` and `longitude` on the record appropriately.

### Actual behavior

Latitude and longitude remain `nil` despite coordinates being defined in the stub.

The `geocode` method relies on calling `latitude` and `longitude` on the result. `Geocorder::Result::Base` defines `latitude` and `longitude` in terms of `coordinates`, but the test result is essentially a hash with no logic to synchronize lat and long with coordinates.

It seems I have to set `coordinates`, `latitude` and `longitude` in stubs to cover all bases. In my own test helpers for stubbing geocoder responses, I've added the following to compensate for this.

```ruby
  def sync_lat_long_and_coordinates(options)
    if options[:coordinates].present?
      options[:latitude] = options[:coordinates].first
      options[:longitude] = options[:coordinates].last
    elsif options[:latitude].present? && options[:longitude].present?
      options[:coordinates] = [options[:latitude], options[:longitude]]
    end

    options
  end
```

### Environment info

* Geocoder version: 1.4.4
* Rails version:  5.1

jheiss:
Looks like this is related to #1197

alexreisner:
Thanks @derekprior and @jheiss. The intention of the Test lookup/result is to be very generic, and to do very little (since all APIs return different data, it's meant to be very flexible). That being said, ALL results return lat/lon based on the `coordinates` array, so I think there's a good case to be made for implementing `latitude` and `longitude` methods in the Test lookup which will keep things in sync. As I think is clear from my odd response on #1197, I thought it was already like this (oops!).

sergiopantoja:
Another temporary workaround 👍 

```ruby
Geocoder::Lookup::Test.add_stub(
  "New York, NY", [
    {
      'coordinates'  => [40.7143528, -74.0059731],
[...]

PR #1311, opened 2018-07-08
Corrected test stub example

sd:
As currently written, the sample stub doesn't work when testing in rails for `geocode` callbacks.

The stub should return separate `'latitude'` and `'longitude'` attributes, instead of `'coordinates'`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -524,7 +524,8 @@ With the above stub defined, any query for "New York, NY" will return the result
     Geocoder::Lookup::Test.set_default_stub(
       [
         {
-          'coordinates'  => [40.7143528, -74.0059731],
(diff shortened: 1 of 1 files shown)


alexreisner:
Please see #1258.

marshallds:
@alexreisner Since the fix won't actually be available until 1.5, would it make sense to merge this to master now and revert it in the 1.5 branch. I thought I was loosing my mind trying to troubleshoot my test suite.

alexreisner:
@marshallds sorry for the delay. I'm actually working on the 1.5 release right now. Should hopefully be out this week.

alexreisner:
@marshallds just released 1.5.0.

PR #1319, opened 2018-07-23
Update README.md to reflect that request.location can be nil.

practicingdev:
I am not entirely sure what causes this to happen, but I experienced the issue in production myself and saw this related StackOverflow link:

https://stackoverflow.com/questions/27114680/geocoder-request-location-returns-nil


alexreisner:
A message about this is already present in [the section about request.location](https://github.com/alexreisner/geocoder/blob/9443e75aeb700f269b92c4b8923701d8de8d8ff0/README.md#geocoding-http-requests).

PR #1322, opened 2018-07-26
Add IP2Location Web Service to IP Address Lookups

ip2location:


alexreisner:
Thanks for this! Please look into test failures. Also, all the methods in [the Result class](https://github.com/ip2location/geocoder/blob/33c18085bb6419c74be0189cfc200db02865287e/lib/geocoder/results/ip2location_api.rb) are redundant as getters are defined dynamically for every item in `response_attributes` (see loop at the bottom of the file).

ip2location:
Hi @alexreisner , 

We have fixed the issue on the Result class. 

However, we are facing problem in the test. Based on the test log, we didn't create test/fixtures/ip2location_api_madison_square_garden because our API is for IP Address Lookup instead of Street Address Lookup.

Could you please guide us on how to pass the test?

Thank you.

alexreisner:
I fixed the tests, made some other changes to the coding style (for consistency with rest of gem) and merged. ([See final merged PR.](https://github.com/alexreisner/geocoder/compare/2267c534d0373e891a7016fb411e465c5943d16b...e329eb36a1a8eb38d2bf2849fea8ca2fcdf95800))

ip2location:
@alexreisner Thank you for your help.

PR #1324, opened 2018-08-01
Add IPInfoDB API

ipinfodb-com:


alexreisner:
Sorry for the delay on this. I'm a little hesitant to merge because IPInfoDB looks like a repackaging of IP2Location. So my question is: why would someone use IPInfoDB over IP2Location? (Reason for asking is that each new API adds a bit of time and effort to maintaining the gem so I'm limiting additions to APIs that offer access to new information, or a different pricing plan, or increased availability, etc.)

ipinfodb-com:
This integration is for IPInfoDB which is free and different from IP2Location

alexreisner:
OK, however there are a bunch of test errors and failures. Could you please take a look at those?

ipinfodb-com:
Hi @alexreisner 

We are facing the branch conflicts issue. May we know how to solve it?

PR #1331, opened 2018-08-23
Allow Esri for_storage to be set on a per call basis

tneems:


alexreisner:
I'm open to adding this feature, but I don't think this does exactly what you want. Current implementation says: set `forStorage = true` if `for_storage` is true in config or in query options. If it's  true in config there's no way to override in options.

tneems:
Good point, it worked for my use case of only passing the options and I hadn't considered what would happen if the configuration was set to true. I amend the commit to always use the options value if one is passed and then to fallback to the confiuration. Was this what you had in mind?

alexreisner:
This is closer, however, we need to be more careful still. The gem currently doesn't set the `forStorage` query string param unless it's explicitly named in the configuration. With your changes, it will always be set (q.s. will have `forStorage=`), even if not configured.

tneems:
If the value isn't configured/`nil` it is being removed by [#url_query_string](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookups/base.rb#L150). In the previous code it was actually impossible to pass `forStorage=false` to Esri, which could potentially cause problems if Esri ever decided to change the default value for `forStorage`(extremely unlikely). That said I'm more than happy to change it to only set `forStorage` is the value is truthy like before.

alexreisner:
Ah, yes, good point. This is great. Thanks for the tests. Merging.

PR #1334, opened 2018-09-04
Allow esri token to be set at call time

tneems:


tneems:
@alexreisner thanks for merging.

alexreisner:
@tneems I'm not sure why Travis didn't run the tests for this--looks like it's causing several errors. Would you mind looking into them?

alexreisner:
@tneems my bad, sorry--never mind!

PR #1339, opened 2018-09-11
Don't try to look up private IP address.

ignatiusreza:
The implementations and results returned from lookup mimic what is done for loopback addresses..

Private ip addresses are received in development mode when the server are run inside docker containers.. doing actual lookup for these ips are a waste of resources, and since the results are generally known, we can optimistically return the expected results..


alexreisner:
Very nice, thanks for this!

ignatiusreza:
my pleasure, thanks for the merge!

PR #1343, opened 2018-09-14
Google Places Details - basic lookup

zjwhitehead:
Due to Google Maps pricing changes we have switched to only requesting basic fields from the API in order to save costs. By default you are charged the maximum amount for all fields.
 
The field breakdown can be viewed here along with pricing for each. https://developers.google.com/maps/billing/understanding-cost-of-use#basic-data

Next on the list would be to support specifying needed fields and only requesting and parsing those in cases where contact or atmosphere data is required.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,7 +13,7 @@ Key features:
 
 * Forward and reverse geocoding, and IP address geocoding.
 * Connects to more than 40 APIs worldwide.
-* Performance-enhancing feaures like caching.
+* Performance-enhancing features like caching.
 * Advanced configuration allows different parameters and APIs to be used in different conditions.
 * Integrates with ActiveRecord and Mongoid.
 * Basic geospatial queries: search within radius (or rectangle, or ring).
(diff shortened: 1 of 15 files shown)


alienxp03:
This is useful. Does it mean right now, all `:google` lookup will be searching for all atmosphere, contact data? We noticed a significant increase in our cost, and it looks like this could be the culprit.

alexreisner:
@alienxp03 This does not apply to the `:google` lookup, only `:google_places_details`.

@zjwhitehead Is there a way to allow users to specify fields without adding a whole new lookup?

zjwhitehead:
> This is useful. Does it mean right now, all `:google` lookup will be searching for all atmosphere, contact data? We noticed a significant increase in our cost, and it looks like this could be the culprit.

Yes @alienxp03 for the default Google lookup you will be paying the max rate regardless of what you fields you use. Our costs jumped a lot too which required this change

zjwhitehead:
> @zjwhitehead Is there a way to allow users to specify fields without adding a whole new lookup?
Thats the ideal change @alexreisner. I put this together for us quickly because we were getting with a significant jump in charges.

Seems like we could just default to the current functionality if an optional param isnt specified in order to maintain backwards compatibility.

Issue #1219, opened 2017-09-10
Sample configuration for ipinfo_io as provider with TOKEN as request param


dikey94:
I'm looking for a sample configuration of Geocoder with ipinfo.io. I know i can set a :ip_lookup => :ipinfo_io but how to put a token?

Thanks

TrangPham:
Hi thanks for bringing this up! It's not noted on the readme for the lookup. Here's the class: https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookups/ipinfo_io.rb

It's expecting it as the API key. 
Please feel free to create a PR to update the read me. :) 

Thanks!

TrangPham:
Beginner-task:

Update the READ ME section for ipinfo.io to include :api-key => TOKEN

PR #1354, opened 2018-10-18
Update IPInfo.io instructions to include api_key usage

rdlugosz:
Resolves issue #1219


alexreisner:
Sorry, but now that I'm looking at #1219, it should really be closed. If we add this example for one lookup then theoretically we should add it for all. But that would add to the character count of an already long and dense README. Since the big README restructuring a few months ago, I think it's sufficiently clear how to do this ([Lookup Config](https://github.com/alexreisner/geocoder#geocoding-service-lookup-configuration) section).

PR #1355, opened 2018-10-18
Allow https for IPInfo.io queries without an API key

rdlugosz:
It appears that at some point they changed the "TLS only for paying
customers" rule.

See: https://ipinfo.io/developers#https-ssl
"Our API is available over a secure HTTPS connection for all users, even
on the free plan."


alexreisner:
Glad to hear this--thanks!

PR #1366, opened 2018-11-30
Fixed typo

ankane:
Fixed a small typo. It shows up when you don't pass SmartyStreets credentials, but doesn't seem to affect the actual geocoding.


alexreisner:
Thanks!

PR #1370, opened 2018-12-04
added street_address functionality to Here api

kyle-s-autobrain:
Added street_address availability to Here api. Simple fix.


alexreisner:
Thanks for this. Unfortunately, street addresses are complicated. Every country does them differently, so my policy is to only provide a `#street_address` method for geocoding APIs that actually return a properly-formatted street address (following the rules of the relevant country). In this case, the method will be correct for the United States but not for most of the rest of the world.

PR #1375, opened 2019-01-10
Add country and mapview as extra parameters to here lookup

jfcaiceo:
Currently Geocoder uses an old version of the `here` API. This change updates the API version and allows the support of country and mapview parameters.


alexreisner:
Thanks for this. What do you think about calling the option `:bounds` instead of `:mapview` for consistency with other lookups? (Assuming I'm understanding its function correctly.)

jfcaiceo:
It's very similar to other `bounds` options, so that's fine with me. I'm going to make the change.

PR #1376, opened 2019-01-13
Allow mixed-case Table name and Column names for ActiveRecord

HashBug:
Quoted Table name and Column names to Allow Uppercase and Mixed Case table/column names.

Example cases:
"address"."LATITUDE"
"ADDRESS"."latitude"
"Address"."Latitude"

### DIFF
--- a/lib/geocoder/stores/active_record.rb
+++ b/lib/geocoder/stores/active_record.rb
@@ -18,19 +18,19 @@ module Geocoder::Store
 
         # scope: geocoded objects
         scope :geocoded, lambda {
-          where("#{table_name}.#{geocoder_options[:latitude]} IS NOT NULL " +
-            "AND #{table_name}.#{geocoder_options[:longitude]} IS NOT NULL")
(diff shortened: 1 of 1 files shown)


alexreisner:
Thanks for this. Unfortunately this change will break certain apps, so it requires a minor release. I'm also not sure what double quotes do on MySQL and SQLite. Given the complexity of the change, and the fact that its purpose is to support a very unusual situation (I assume the intention is for Postgres columns that were created case-sensitively, with double quotes), I don't think it will be merged. Sorry.

HashBug:
You guessed it right, the issue is with the PG columns as the table and column names are all upper case (Imported from a PHP application). Batch geocoding fails complaining about the table/column name. 
Looking at the activerecord's implementation of **where** query, I thought quoting the table and column names will work for all scenarios.
Will wait for the future release to fix this. For now I have done my own rake implementation for batch geocoding.
Thanks for taking time to review this.

Issue #1378, opened 2019-01-16
 By default Geocoder will rescue any exceptions raised by calls to a geocoding service and return an empty array


meetme2meat:
According to [README](https://github.com/alexreisner/geocoder/blob/master/README.md#error-handling) 
> By default, Geocoder will rescue any exceptions raised by calls to a geocoding service and return an empty array

Well, this holds mostly true for all exception except the [`Geocoder::NetworkError`](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookups/base.rb#L311). It [seems like](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookups/base.rb#L196-L202)(no rescue for **Geocoder::NetworkError**) system would raises `Geocoder::NetworkError` instead of returning an empty array in response which break the README.md declaration.

### Expected behavior
  Geocoder.search would raise `Geocoder::NetworkError`.

### Actual behavior
   Geocoder.search api should return [] (an empty array)
  

### Environment info

* Geocoder version:  1.5.0
* Rails version: 5.0.0.1

alexreisner:
Good catch. Yes, this should be fixed.

philosoralphter:
Any idea in what version this was introduced?

dsandstrom:
@philosoralphter Looks to be added in v1.5.2 https://github.com/alexreisner/geocoder/commits/v.1.5.2

PR #1380, opened 2019-01-21
Rescue the Geocoder::NetworkError to stay consistent with the README.md

meetme2meat:
Closes https://github.com/alexreisner/geocoder/issues/1378

### DIFF
--- a/lib/geocoder/lookups/base.rb
+++ b/lib/geocoder/lookups/base.rb
@@ -197,6 +197,8 @@ module Geocoder
         raise_error(err) or Geocoder.log(:warn, "Geocoding API connection cannot be established.")
       rescue Errno::ECONNREFUSED => err
         raise_error(err) or Geocoder.log(:warn, "Geocoding API connection refused.")
+      rescue Geocoder::NetworkError => err
+        raise_error(err) or Geocoder.log(:warn, "Geocoding API connection is either unreacheable or reset by the peer")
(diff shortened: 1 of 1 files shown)


alexreisner:
Thanks for this! The lines you added are good, however I think they should be further down in the file, near line 308. And we might as well do the same thing for LookupTimeout. Does that make sense?

meetme2meat:
@alexreisner LookupTimeout [inherit from](https://github.com/alexreisner/geocoder/blob/1adf2e520a0c568a046e8ef70bed98c5be2038a2/lib/geocoder/exceptions.rb#L34) **Timeout::Error** so in a sesnse the LookupTimeout is essentially rescue over [here](https://github.com/alexreisner/geocoder/blob/f9d16c4aababc9b3c64966e294dda4fc9d260722/lib/geocoder/lookups/base.rb#L202).

> The lines you added are good, however I think they should be further down in the file near line 308.

So you referring over here is do not raise the `Geocoder::NetworkError`  line 311 https://github.com/alexreisner/geocoder/blob/1d7d53690ce2c32c1b8a7ea87a6e72bcb13fdae1/lib/geocoder/lookups/base.rb#L311

and have this instead
```
raise_error(Geocoder::NetworkError) or Geocoder.log(:warn, "Geocoding API connection is either unreachable or reset by the peer") 
```

PR #1382, opened 2019-02-05
Add adjustment to freegeoip to use freegeoip.app

mixergtz:
Since freegeoip was deprecated a similar service called freegeoip.app provides the same results, the only thing mandatory is to use `https`

This PR makes adjustments to work with `:freegeoip` as before without changes for the end user

I was thinking about creating a new API provider but since freegeoip is deprecated, simply modifying it could bring it to life again

The same can be achieved without this PR, using:
```
Geocoder.configure(
  ip_lookup: :freegeoip,
  freegeoip: { host: "freegeoip.app" },
  use_https: true
)
```

### DIFF
--- a/lib/geocoder/lookups/freegeoip.rb
+++ b/lib/geocoder/lookups/freegeoip.rb
@@ -10,7 +10,7 @@ module Geocoder::Lookup
 
     def supported_protocols
       if configuration[:host]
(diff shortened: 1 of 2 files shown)


alexreisner:
Great pull request--thanks!

PR #1383, opened 2019-02-14
Implementing autocomplete url to here and google_places_search

MrFenril:
Hi !
As you may know, some API also offer addresses autocompletion. I tought that it could be interesting for other people to have this feature in the project as I also need it myself.
For now I did it only for Here and Google_places_search lookups, but I can implement others if needed. 
Let me know if you're interested or if you have some feedback about the code (:.

### DIFF
--- a/lib/geocoder/lookups/base.rb
+++ b/lib/geocoder/lookups/base.rb
@@ -76,6 +76,8 @@ module Geocoder
       # subclss this method, they must also subclass #cache_key.
       #
       def query_url(query)
+        return autocomplete_query_url + url_query_string(query) if query.complete?
+
         base_query_url(query) + url_query_string(query)
       end
 
@@ -106,6 +108,10 @@ module Geocoder
       #
       def base_query_url(query)
         fail
+      end
+
+      def autocomplete_query_url
+        raise NotImplementedError
       end
 
       ##
--- a/lib/geocoder/lookups/google.rb
+++ b/lib/geocoder/lookups/google.rb
@@ -47,7 +47,7 @@ module Geocoder::Lookup
     def results(query)
       return [] unless doc = fetch_data(query)
       case doc['status']; when "OK" # OK status implies >0 results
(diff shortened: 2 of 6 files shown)


alexreisner:
Thanks for this! Very interesting idea. However, unfortunately, I think it's just a bit outside the intended scope of Geocoder. I'm trying to be careful about keeping it focused, so that maintenance doesn't get out of hand. Sorry about that!

fab-girard:
Hello,
first, thanks for your very quick answer. It's a please.
Unfortunatly, I'm not sure today the result of geocoding has a sense without autocompletion feature... today all relevant services for geocoding feature the ability to autocomplete the text user is currently entering... it's now like a standard. Thinks about a web or mobile app today without this feature. So another point of view could be autocomplete is fully part of geocoder service : it's in fact a guide with goal getting lat/lng.
Anyway to reconsider your opinion about this feature?
Fab

alexreisner:
Thanks, this is a good point. I'm thinking about it... Do you know which other APIs offer autocompletion?

fab-girard:
Hello,
thanks a lot, we appreciate.
Those services have for instance a specific endpoint for complete/suggest: Google, Here, Mapbox, ESRI/ArcGIS, etc...
Other services have no specific endpoint but they have a parameter in their geocode operation to "tune" it for autocompletion, it's the case for Addok french geocoder (called "BAN data gouv" in this gem).
Fabien

alexreisner:
One of my main concerns is that some APIs might require different credentials for accessing different services (geocode vs autocomplete), which would require changes to how API keys are configured. That is, unless the autocomplete requests were handled by separate lookups, but then the existing code for connecting to APIs would have to be module-ized, which means a change in architecture...

Basically, this would be nice to have, but it's not a core feature of Geocoder, so if the implementation complicates use or adds significant code complexity, I don't think we should add it. I'm open to ideas, though.

PR #1386, opened 2019-03-07
Add Netherlands National Geocode Register

opensourceame:
This PR adds support for the Netherlands National Geocode Register which is considered the most accurate address register in NL

https://geodata.nationaalgeoregister.nl/

No API key is required for use of the service.


opensourceame:
Can anyone help with this PR? The Travis tests are failing but for expected reasons, for example:

1. The .nl API always returns a result nomatter what the search term is. Searching for "no results" as a search term will always return results therefore the "no results" test is not valid in context of the .nl API.

2. The .nl API does not support lookups by coordinates, only text search terms.

3. The .nl API only returns data relevant to the Netherlands therefore the "Madison Square Garden" search test will not return the expected coordinates in the test.

Is there a way to turn off the tests that are not relevant @alexreisner ?

alexreisner:
Thanks, and sorry for the slow response on this. I disabled the irrelevant tests, fixed one minor bug, and merged. However, I can't read Dutch so it would be helpful if someone could help me fill in the missing data in the [README API GUIDE](https://github.com/alexreisner/geocoder/blob/master/README_API_GUIDE.md#nationaal-georegister-netherlands-nationaal_georegister_nl).

PR #1397, opened 2019-04-17
Add support for Ordnance Survey Names API

pezholio:
This PR adds support for the [Ordnance Survey Names API](https://apidocs.os.uk/docs/os-names-overview), which supports lookup of named places (such as towns, cities, villages etc) and postcodes in the UK. 

Results are given in Easting / Northing, so I've had to add a [gem dependency](https://github.com/theodi/breasal), which does the conversion. Hope this is OK.


alexreisner:
Unfortunately, Geocoder has no gem dependencies currently, and it's a sort of unwritten policy not to introduce them. Is there a way around that?

pezholio:
OK, that’s not a blocker. I can easily extract the relevant code into a class within Geocoder (it’s just maths). I wrote the original gem, and the code could use a tidy anyway. I’ll see what I can do 👍🏻

pezholio:
I've added a new class now, called `EastingNorthing` that I've adpated from the original code in the Breasal gem, and removed the dependency.

alexreisner:
Thanks for the changes, and very sorry about the delay on this. A few comments:

* I think the name of this lookup should be `:uk_ordnance_survey_names`. For one thing, it's a geographically-limited lookup and stating that in the name seems helpful. For another, this will avoid confusion with the (admittedly, poorly-named) `:osmnames` lookup.
* `gemspec path: '../'` is environment-specific and should be removed from all Gemfiles.
* In the `Result` class, it's unnecessary to define `latitude` and `longitude` methods.

Other than that, this looks good to merge.

pezholio:
@alexreisner These should all be addressed now! 👍

alexreisner:
Thanks! Just fixed up one tiny thing: 5c41aa888cc21b1f458320932118ce4ffec22333

pezholio:
Ah, course! That works much better, thanks 😊

PR #1398, opened 2019-04-22
Correct param name for api key is "apikey"

HatlamMe:
https://tech.yandex.ru/maps/doc/geocoder/desc/concepts/input_params-docpage/


olegkozyrev:
As far as I know it's `apikey` instead of `api_key`

Struggled myself today with that :)

HatlamMe:
Yep. Shame on me

PR #1406, opened 2019-08-19
Add support for Ipregistry

lpellegr:
We have users who would like to use your great Ruby library with [Ipregistry](https://ipregistry.co). This PR adds support for this new service that is reliable and returns responses with very low latency worldwide.

### DIFF
--- a/README_API_GUIDE.md
+++ b/README_API_GUIDE.md
@@ -341,6 +341,16 @@ Data Science Toolkit provides an API whose response format is like Google's but 
 IP Address Lookups
 ------------------
 
+### Ipregistry (`:ipregistry`)
+
+* **API key**: required (see https://ipregistry.co)
+* **Quota**: first 100,000 requests are free, then you pay per request (see https://ipregistry.co/pricing)
+* **Region**: world
+* **SSL support**: yes
+* **Languages**: English
+* **Documentation**: https://ipregistry.co/docs
+* **Terms of Service**: https://ipregistry.co/terms
+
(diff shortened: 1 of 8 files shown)


lpellegr:
Some test jobs are failing but failures are not related to the changes introduced by this PR. The failures look similar to the ones currently on the master.

jmarbach:
Really thrilled to see this! Ipregistry offers much needed supplemental context that is useful for a wide range of use cases.

lpellegr:
@alexreisner Is this repository still maintained?

lpellegr:
@alexreisner Thanks for your answer and the review.

> in lookups/ipregistry.rb: If the only valid host is api.ipregistry.co, we should not look at a user-configured value for :host.

For some users, we provide a docker image they deploy by themselves. In that case, they need to specify their own value. Your comment let me noticed I missed HTTP as a supported protocol. I have added it to the list of supported protocols.

> in results/ipregistry.rb: What's the purpose of flattening the hash? Why not use the accessor methods to locate the data in the structure?

It prevents several `nil` checks and touching existing tests. Furthermore, it looks like existing integrations use a flat hierarchy so it's a way to keep this uniform.

> in README_API_GUIDE.md: Please insert the new lookup at the bottom of the list, rather than at the top.

Done.

alexreisner:
OK, thanks. Merging.

lpellegr:
@alexreisner Thanks! Is there any plan for a release soon?

alexreisner:
Yes, probably this month.

PR #1408, opened 2019-08-30
Add mapkit geocoder

eilers:
Hi!
I needed to switch from Google to Apple-Maps. Since Apple provides MapKit-JS, we can use it for Geocoding as well. :)

Hope you will find it worth to merge. 

Regards, Stefan

### DIFF
--- a/README_API_GUIDE.md
+++ b/README_API_GUIDE.md
@@ -546,4 +546,23 @@ You must add either the *[hive_geoip2](https://rubygems.org/gems/hive_geoip2)* g
     )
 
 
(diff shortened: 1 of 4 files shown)


alexreisner:
Sorry, but I'm closing this because I suspect it violates Apple's terms of service. From the [documentation](https://developer.apple.com/documentation/mapkitjs/mapkit/geocoder): "When using MapKit JS to show Apple Maps on a website, the geocoder handles forward and reverse geocoding." I think it's only meant to be used by websites that are displaying Apple Maps. If we can find a definitive statement to the contrary, I'd be happy to consider this.

PR #1411, opened 2019-09-29
fix yandex language param

acasaism:
The **plng** parameter didn't change the response language in yandex. Look this: 
https://tech.yandex.com/maps/geocoder/doc/desc/concepts/input_params-docpage/
With the **lang** parameter we fix the problem


alexreisner:
Thanks!

PR #1416, opened 2019-10-10
Feature/allow coordinates in ban fr

Mziserman:
I was not able to complete a bundle install (it has been resolving dependencies for a day now). 

Or start tests as 
```bash
ruby test/unit/lookups/ban_data_gouv_fr_test.rb
```
results in
```bash
	2: from test/unit/lookups/ban_data_gouv_fr_test.rb:2:in `<main>'
	1: from ~/.rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require'
~/.rbenv/versions/2.6.3/lib/ruby/2.6.0/rubygems/core_ext/kernel_require.rb:54:in `require': cannot load such file -- test_helper (LoadError)
```

what am I doing wrong ?

I was however able to load geocoder in irb and test the methods I used.

### DIFF
--- a/lib/geocoder/lookups/ban_data_gouv_fr.rb
+++ b/lib/geocoder/lookups/ban_data_gouv_fr.rb
@@ -86,6 +86,12 @@ module Geocoder::Lookup
       unless (citycode = query.options[:citycode]).nil? || !code_param_is_valid?(citycode)
         params[:citycode] = citycode.to_s
       end
+      unless (lat = query.options[:lat]).nil? || !latitude_is_valid?(lat)
+        params[:lat] = lat
+      end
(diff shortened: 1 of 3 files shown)


alexreisner:
Thanks for this, and sorry for the slow response. You should be able to run tests locally using `rake test`. When I run them I get one failure:

    <"http://api-adresse.data.gouv.fr/search/?lat=48.789&lon=2.789&q=13+rue+yves+toudic%2C+75010+Paris"> expected but was
    <"https://api-adresse.data.gouv.fr/search/?lat=48.789&lon=2.789&q=13+rue+yves+toudic%2C+75010+Paris">

The only difference is "http" and "https". If you fix that, I'll merge.

PR #1418, opened 2019-10-17
add apikey to yandex base_query_url

IgorKamrakov91:


alexreisner:
Thanks for this. What problem is this intended to solve?

IgorKamrakov91:
@alexreisner sorry for not getting back to you sooner, hope its still relevant. 
yandex changed the rules for using service. now with request you need to send api key. so i made fork with some changes and thought pr to main project will be helpfull too)

alexreisner:
Ah, that makes sense. However, Geocoder already passes the `apikey` if you set it in the configuration. So I think this change is unnecessary.

PR #1419, opened 2019-10-23
Update IPinfo quota information

m-an:
Source:
https://ipinfo.io/missingauth
Screenshot_2019-10-23 Get the full potential of IPinfo - IPinfo io

I've got confused having a rate limit of 1k/day, so maybe this will be clearer having it mentioned here.


alexreisner:
Thanks!

Issue #1429, opened 2019-12-12
"Here" introduced new urls and a single API key


stephanpavlovic:
It looks like "Here" is using new urls and also a new authentication method, I have found no informations about when they will remove the legacy url but fo new REST/JS Apps you can not even generate an app code anymore....

https://developer.here.com/documentation/geocoder/topics/dev_guide/request-constructing.html

https://developer.here.com/documentation/authentication/dev_guide/topics/api-key-credentials.html

alexreisner:
Thanks! Definitely open to a pull request, if anyone wants to handle this.

stephanpavlovic:
I will try to handle this

stephanpavlovic:
I think I will provide something one the weekend.
One quick question @alexreisner: I guess I nee to make it work in parallel for both, the old and the new way. Is this also the case in one of the other lookup services, so I can have a look how you handle such a case there?

alexreisner:
Thanks, and good question, but I don't think it needs to work both ways. If the new URL and key format are working, we can just remove support for the old format.

alexreisner:
Fixed by #1430.

PR #1430, opened 2019-12-13
Use the Here lookup with the new Here url and the new authentication method

stephanpavlovic:


alexreisner:
This looks good, but as I just mentioned in #1429, I don't think there's any need to keep the `app_id` stuff around. (If someone hasn't switched to the new format yet, they can stick with a Geocoder version that supports the old format.)

stephanpavlovic:
Alright, I removed the app_id stuff

alexreisner:
Thanks! Looks great. Merged.

PR #1437, opened 2020-01-14
Remove bad information from HERE API docs

valscion:
Seems like with https://github.com/alexreisner/geocoder/issues/1429 and https://github.com/alexreisner/geocoder/pull/1430, the API key for HERE integration should now be provided like so:

```
Geocoder.configure(api_key:  "my api key")
```

The API key is displayed in HERE Project Page as documented here: https://developer.here.com/documentation/authentication/dev_guide/topics/api-key-credentials.html

Pasted_Image_14_1_2020__14_31


alexreisner:
Yes, good catch!

PR #1446, opened 2020-04-09
[fix] ruby 2.7.1 warning

skcc321:
lib/geocoder/lookups/latlon.rb:28: warning: `elsif' at the end of line without an expression


alexreisner:
Excellent, thanks!

Issue #1396, opened 2019-03-28
 IP Address Not Detected for IPv6 with Port


vbyno:
### Expected behavior

`Geocoder.search` returns correct `Geocoder::Result::Ipstack` results

### Actual behavior

`Geocoder.search` returns incorrect `Geocoder::Result::Bing` results

### Steps to reproduce
1) set `bing` lookup and `ipstack` as `Ip_lookup`
```
Geocoder.configure(:timeout=>15, :lookup=>:bing, :ip_lookup=>:ipstack, :language=>:en, :http_headers=>{}, :use_https=>false, :http_proxy=>nil, :https_proxy=>nil, :api_key=>nil, :basic_auth=>{}, :logger=>:kernel, :kernel_logger_level=>2, :always_raise=>:all, :units=>:mi, :distances=>:linear, :ipstack=>{:api_key=>"xxxxxx"}, :here=>{:api_key=>["xxxxxx", "xxxxxx"]}, :bing=>{:api_key=>"xxxxxx"})
```
2) Run: `Geocoder.search("[2607:fea8:1360:f7d:dce7:b7f0:e0b6:1014]:80")`
Results are:
```
[...]

PR #1447, opened 2020-04-14
Support IP search with port

gdomingu:
Resolves issue where trying to search via an IP address with a port was not working properly

Example:
`Geocoder.search("[3ffe:0b00:000:0000:0001:0000:0000:000a]:80")`

Resolves https://github.com/alexreisner/geocoder/issues/1396


alexreisner:
Excellent. Thank you!

PR #1448, opened 2020-04-22
Yandex fix. No more "found" param in meta section

HatlamMe:
Yandex response has changed. 
There was param named "found" before. But it was removed. So any geocode request returned empty array.
Actually i consider the condition as redundant.

### DIFF
--- a/lib/geocoder/lookups/yandex.rb
+++ b/lib/geocoder/lookups/yandex.rb
@@ -34,7 +34,7 @@ module Geocoder::Lookup
       end
       if doc = doc['response']['GeoObjectCollection']
         meta = doc['metaDataProperty']['GeocoderResponseMetaData']
-        return meta['found'].to_i > 0 ? doc['featureMember'] : []
+        return doc['featureMember'].to_a
       else
         Geocoder.log(:warn, "Yandex Geocoding API error: unexpected response format.")
         return []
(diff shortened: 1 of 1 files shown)


alexreisner:
Thanks for this. Do you have a link to the Yandex documentation for this new format?

HatlamMe:
Sorry,  no documentation. It’s just got broken a couple of hours before. And I’ve made a hot fix. Actually it’s pretty old API, it should not change. Here is the link to the doc: 
https://tech.yandex.ru/maps/geocoder/doc/desc/reference/response_structure-docpage/ (Russian)
But it’s pretty old, and it still has “found”.

alexreisner:
Thanks. In that case, would it be possible to update the sample JSON in `test/fixtures/yandex_...` to reflect actual API responses? Would be good to keep all parts of the code consistent.

HatlamMe:
It looks like it was a bug at Yandex, now "found" came back in response. But I would propose to accept my pull request — in any case, at least it wont get worse as the condition was not necessary.

alexreisner:
Yes, I agree with that. Do you want to also remove the line above the one you changed, where `meta` is assigned? It now seems unnecessary.

HatlamMe:
Sure! Sorry didn't notice in a hurry. Can you please smash 2 commits while merging?

alexreisner:
No problem. Thanks!

PR #1450, opened 2020-04-29
Update from Mashape to RapidAPI

alexwalling:
The Mashape API proxy has been deprecated and needs to be replaced with RapidAPI proxy URL to work reliably.

### DIFF
--- a/README_API_GUIDE.md
+++ b/README_API_GUIDE.md
@@ -421,7 +421,7 @@ IP Address Lookups
 * **Region**: world
 * **SSL support**: yes
 * **Languages**: English
-* **Documentation**: https://market.mashape.com/fcambus/telize
+* **Documentation**: https://rapidapi.com/fcambus/api/telize
 * **Terms of Service**: ?
 * **Limitations**: ?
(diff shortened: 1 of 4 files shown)


timrwilliams:
To add some more detail this change is required to fix the hosted Telize lookup service which is currently broken for all versions of this gem. 

We're currently having to remediate this via a monkey patch in our initializer:  

```
#config/initializers/geocoder.rb
Geocoder::Lookup::Telize.class_eval do
  def query_url(query)
    "#{protocol}://telize-v1.p.rapidapi.com/location/#{query.sanitized_text}?rapidapi-key=#{api_key}"
  end
end
```

alexreisner:
Thanks for this (and sorry for the Hound auto-comments--I've yet to configure it). The code looks good and I'd like to get this merged. The only obstacle is getting an API key so I can test it. I tried the [contact form](https://rapidapi.com/contact-us) on RapidAPI.com but it wouldn't submit ("Submission failed, please try again later."). Any thoughts on how I might run a few test queries from my local machine? Is there someone I can contact directly?

timrwilliams:
I've sent you one to use via the contact form on your website.

alexwalling:
If you have any issues with the key sent to you by Tim, just shoot an email to support@rapidapi.com and they can help you there.

alexreisner:
The key worked great. Thanks!

PR #1452, opened 2020-05-19
Updated Geocodio API version to 1.6

MiniCodeMonkey:
See changelog here: https://www.geocod.io/docs/#changelog


alexreisner:
Excellent, thanks for the update. Given that there are changes to response structure between version 1.3 and 1.5 (eg, `formatted_address` has new key, according to changelog you link to above), it seems like we should update the sample responses in `test/fixtures/geocodio_*`.

MiniCodeMonkey:
@alexreisner Thanks for the prompt response! I apologize for the delay in my end. In the meantime we just pushed API v1.6 in rapid succession so I have updated this PR to reflect that.

I have updated fixtures and test cases to reflect the newest API output. The `formatted_address` output is only different if a secondary address unit is supplied, so in this case these test cases do not need to be changed at this time.

Thanks!

alexreisner:
Perfect. Thanks!

PR #1454, opened 2020-06-03
Support state_code in results from db-ip.com

adamalbrecht:
Currently, the `state_code` method on a result from db-ip.com returns the full state name. But the data returned from their API actually does include a 2 character state code, under `stateProvCode`.

Here is a sample query that I just performed:

```irb
irb(main):005:0> results = Geocoder.search("96.27.101.109")
irb(main):006:0> results[0].state_code
=> "Ohio"
irb(main):007:0> results[0].data['stateProvCode']
=> "OH"
```

### DIFF
--- a/lib/geocoder/results/db_ip_com.rb
+++ b/lib/geocoder/results/db_ip_com.rb
@@ -16,7 +16,7 @@ module Geocoder::Result
     end
 
     def state_code
-      @data['stateProv']
+      @data['stateProvCode']
     end
     alias_method :state, :state_code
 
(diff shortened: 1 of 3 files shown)


alexreisner:
Thanks for this! The weird thing is that I can't find any mention of `stateProvCode` in their documentation. I'm going to write to them to see if I can get an API key to try it out. Which plan are you using?

adamalbrecht:
I noticed the same lack of documentation, but it's definitely there.

Our account page says that we have a "Lite with Basic API" account.

alexreisner:
From customer support: "The stateProvCode response field is indeed a recent addition to our API services and it will be added to the documentation with our next website update. When available the field is populated with the state or province ISO 3166-2 code."

Thanks @adamalbrecht.

PR #1459, opened 2020-07-01
Remove Hash extensions

dduugg:
This PR removes `HashRecursiveMerge` as a core extension to `Hash`. The `HashRecursiveMerge` methods seem to have a single [call site](https://github.com/alexreisner/geocoder/blob/b2f6fb35/lib/geocoder/configuration.rb#L88), invoked by a `Geocoder::ConfigurationHash`, which already [includes](https://github.com/alexreisner/geocoder/blob/b2f6fb35/lib/geocoder/configuration_hash.rb#L5) the module.

Monkey-patching the core library can cause problems, such as unintentionally coupling code to the `geocoder` library (which can break in unanticipated ways when code is refactored), or causing conflicts with other libraries that have conflicting core extensions.

I'd also like to move `HashRecursiveMerge` under the `Geocoder` namespace, let me know if you'd be open to such a change.


alexreisner:
Totally agree. The current solution is too heavy. All that's really needed is a utility method that takes two hashes and returns their merged result (which isn't very OOPish, but when it comes to merging two objects I'm not convinced that one of the objects should be responsible for the process). I'm up for redesigning this whole thing. Do you have an overall plan in mind?

dduugg:
Yeah, sounds good. I could create `lib/geocoder/util.rb` with an `rmerge!(h1, h2)` method, if that works for you. Let me know if you have other ideas.

alexreisner:
Sounds good to me!

alexreisner:
This looks good! Just a couple of things:

1. Since the `rmerge!` method has lost the context of a hash object, we should probably give it a more descriptive name. Also I think the `!` is unnecessary. What about `recursive_hash_merge`?
2. The test seems to be failing:

    ...geocoder/test/unit/util_test.rb:9:in `test_rmerge!'
    Error: test_rmerge!(UtilTest): NameError: uninitialized constant UtilTest::Util

dduugg:
Thanks, I've made the suggested changes (it wasn't meant to reviewed juuust yet, but I appreciate it nonetheless – i was just pushing changes to trigger tests, since I'm having issues running them locally).

lib/geocoder/configuration_hash.rb
```diff
@@ -1,11 +1,11 @@
...
-
 module Geocoder
   class ConfigurationHash < Hash
-    include HashRecursiveMerge
-
     def method_missing(meth, *args, &block)
       has_key?(meth) ? self[meth] : super
     end
+
+    def respond_to_missing?(meth, include_private = false)
```
dduugg, line 7: there's some unrelated work here to placate my editor's linter – lmk if you'd prefer to leave them out.

lib/hash_recursive_merge.rb
```diff
@@ -1,73 +0,0 @@
...
-# Merges a Ruby Hash recursively, Also known as deep merge.
-# Recursive version of Hash#merge and Hash#merge!.
-# 
-# Category::    Ruby
-# Package::     Hash
-# Author::      Simone Carletti <weppos@weppos.net>
-# Copyright::   2007-2008 The Authors
-# License::     MIT License
-# Link::        http://www.simonecarletti.com/
-# Source::      http://gist.github.com/gists/6391/
```
dduugg, line 13: i've intentionally left out this header info (the gist link 404s now anyway), lmk if you'd prefer to include it

alexreisner:
Looks great. Thanks!!

Issue #1460, opened 2020-07-07
Google places search cost cannot be controlled


maximilientyc:
Google has an API endpoint for its Place search services which allows you to control which fields are going to be available in the response. Using this restriction allows you to control the price of API calls made to the Places API.

This gem uses the [Text search requests](https://developers.google.com/places/web-service/search#TextSearchRequests) for the `google_places_search` component. Here is what Google is saying about this endpoint: 

```
Nearby Search and Text Search return all of the available data fields for the selected place 
(a subset of the supported fields), and you will be billed accordingly.
There is no way to constrain Nearby Search or Text Search to only return specific fields. 
To keep from requesting (and paying for) data that you don't need, use a Find Place request instead.
```

I would suggest we migrate to using the [Find place requests](https://developers.google.com/places/web-service/search#FindPlaceRequests) endpoint, so we can specify which data we want to have in the response given by Google, and as a consequence only pay for what we need.

**Behavior proposal**

The component should request for the following parameters to be included in the response: `business_status, formatted_address, geometry, icon,name, permanently_closed, photos, place_id, plus_code, types`. This will exclude fields coming from `Atmosphere Data` and `Contact Data`.

You need to be able to override the default values for the `fields` param, by giving for example: `formatted_address,open_now` which combines data coming from `Basic Data` and `Contact Data`.

For the detailed pricing visit this page: https://cloud.google.com/maps-platform/pricing/sheet/?_ga=2.64143597.318188449.1594045771-1706466804.1555340798

alexreisner:
Thanks, and yes, I agree with this approach. Definitely open to a pull request.

maximilientyc:
I’m going to start working on something, will keep you updated using this thread 👌

maximilientyc:
is being fixed through: https://github.com/alexreisner/geocoder/pull/1461

zjwhitehead:
This is nice and flexible. We did a more hardcoded approach originally 
https://github.com/rvshare/geocoder/commit/3e6273afceb1535718e08955df379c94efa747ef#diff-5a1e50427808e6aade11666cf82b1dceR7

alexreisner:
Yes, thanks again for the good work @maximilientyc!

PR #1461, opened 2020-07-16
Control Google places search API price

maximilientyc:
References issue: #1460

### DIFF
--- a/lib/geocoder/lookups/google_places_search.rb
+++ b/lib/geocoder/lookups/google_places_search.rb
@@ -18,8 +18,8 @@ module Geocoder
 
       private
 
-      def results(query)
-        super(query, 'candidates')
+      def result_root_attr
+        'candidates'
       end
 
       def base_query_url(query)
--- a/lib/geocoder/lookups/google_places_search.rb
+++ b/lib/geocoder/lookups/google_places_search.rb
@@ -18,15 +18,41 @@ module Geocoder
 
       private
 
+      def results(query)
+        super(query, 'candidates')
+      end
+
       def base_query_url(query)
-        "#{protocol}://maps.googleapis.com/maps/api/place/textsearch/json?"
+        "#{protocol}://maps.googleapis.com/maps/api/place/findplacefromtext/json?"
       end
 
       def query_url_google_params(query)
         {
-          query: query.text,
+          input: query.text,
+          inputtype: 'textquery',
+          fields: fields(query),
           language: query.language || configuration.language
         }
+      end
+
+      def fields(query)
+        query_fields = query.options[:fields]
+        return format_fields(query_fields) if query_fields
+
+        default_fields
+      end
+
+      def default_fields
+        legacy = %w[id reference]
+        basic = %w[business_status formatted_address geometry icon name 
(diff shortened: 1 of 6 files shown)


lib/geocoder/lookups/google_places_search.rb
```diff
@@ -18,16 +18,42 @@ def supported_protocols
...
         {
-          query: query.text,
+          input: query.text,
+          inputtype: 'textquery',
+          fields: fields(query),
           language: query.language || configuration.language
         }
       end
+
+      def fields(query)
```
maximilientyc, line 38: By default every fields are requested, to prevent any regression from developers already using the gem.

lib/geocoder/lookups/google_places_search.rb
```diff
@@ -18,16 +18,42 @@ def supported_protocols
...
+        query_fields = query.options[:fields]
+        return format_fields(query_fields) if query_fields
+
+        default_fields
+      end
+
+      def default_fields
+        legacy = %w[id reference]
```
maximilientyc, line 46: `id` and `reference` fields were already available in the previous API response (Text search response), I added those 2 fields here to prevent any regression for developers already using the gem

alexreisner:
This looks great! Thanks for preserving the existing behavior, even if it's not the most cost-effective. My only question is whether we need to add the `root_attr` argument to the `results` method. It doesn't seem to be used, and I'd rather not change the method signature if we can avoid it.

maximilientyc:
It's actually being used in the method body, here: https://github.com/alexreisner/geocoder/pull/1461/files#diff-ec56420a19546fb7982cb25cfcc5bc52R51.
I think there's something to be refactored with how the results from Google are handled for each component (place search, details and premium). For this PR I had the idea to touch only the Place search component, and in another PR work on the refactoring of the Google clients.
Let me know what you think.

alexreisner:
Sorry, my comment wasn't clear. What I don't see is any code that calls `results` with a second argument. In other words, if the line you linked to above was changed to `return doc['results']` would it break anything?

As for separate refactoring PR, sounds good. And if that PR would take advantage of this second `results` argument, then that's where the argument should be added.

maximilientyc:
Thanks for the clarification 👍 

After quickly checking the components inheriting from `Google` class, each component is implementing a different root attr:

|Component|Root attr|
|--|--|
|`GooglePremier`|`results`|
|`GooglePlacesDetails`|`result`|
|`GooglePlacesSearch`|`candidates`|

The `results` method is defined twice, I think we should change that. Once in the `Google` class, once in `GooglePlacesDetails < Google`. A quick and efficient refactor to be done in that PR would be to have a second parameter to the `result` method `root_attr` as a mandatory parameter. Every component inheriting from `Google` will have to implement the following result method, for example:

```ruby
def results(query)
  super(query, 'candidates')
end
```

alexreisner:
That sounds like a good plan to me!

maximilientyc:
I just made the changes, it's ready for review again. Don't hesitate if anything is unclear.

alexreisner:
Perfect! Thanks for the good work on this.

PR #1462, opened 2020-07-22
Add support for Geoapify

ahukkanen:
This adds support for [Geoapify](https://www.geoapify.com) Geocoding API.

PR #1464, opened 2020-07-22
Update Pelias documentation

ahukkanen:
Some of the links in the Pelias documentation are not up to date.

This updates them to refer to the up to date documentation available at:
https://github.com/pelias/documentation

This also adds information about a couple of service providers for managed instances I found. There doesn't seem to be any references available to any other service providers than the "official" one (Geocode Earth). I just accidentally noticed Geoapify is also providing such service when implementing #1462 (their cloud service is not Pelias compatible but they provide managed Pelias as a separate service). I think these deserve to be listed somewhere for easier discovery.

### DIFF
--- a/README_API_GUIDE.md
+++ b/README_API_GUIDE.md
@@ -187,15 +187,20 @@ The [Google Places Search API](https://developers.google.com/places/web-service/
 
 ### Pelias (`:pelias`)
 
+Open source geocoding engine which can be self-hosted. There are multiple service providers that can host Pelias instances (see notes).
+
 * **API key**: configurable (self-hosted service)
 * **Quota**: none (self-hosted service)
 * **Region**: world
 * **SSL support**: yes
-* **Languages**: en; see https://mapzen.com/documentation/search/language-codes/
-* **Documentation**: http://pelias.io/
-* **Terms of Service**: http://pelias.io/data_licenses.html
-* **Limitations**: See terms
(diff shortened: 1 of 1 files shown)


orangejulius:
Hi! Pelias core maintainer here (and co-founder of Geocode Earth). Figured I'd chime in here even though this PR is almost a year old.

All the information mentioned in this PR is good. The links to Pelias documentation are correct and much more relevant than the link to the general Pelias website.

Personally I'd recommend removing all mention of any Pelias-based service providers (even Geocode Earth). It makes more sense to add service providers as their own entry as you've done in some of your other PRs. Especially if, as you've mentioned, they don't actually provide a Pelias-compatible public API.

I'm happy to open a new PR with those changes if it would be merged and the original contributor of this PR isn't around any more :)

alexreisner:
@orangejulius thanks for this. I'm going to merge this PR and then why don't you submit another one with whatever changes you'd recommend? I don't know how this PR got overlooked. (@ahukkanen sorry!)

ahukkanen:
Thanks for merging @alexreisner !

Regarding your point @orangejulius 

> Especially if, as you've mentioned, they don't actually provide a Pelias-compatible public API.

They do provide a default Pelias installation which is naturally Pelias compatible. Their own geocoding API is just not Pelias compatible, although I believe it has drawn some inspiration from it being quite similar. So, they are two separate things.

PR #1472, opened 2020-10-26
Reverse coordinates when parsing from NationaalregisterNl

Marthyn:
The coordinates are returned in a specific notation from the NationaalRegisterNl API:

`POINT(4.89133064 52.3739109)`

The result class parses that so we get 2 individual coordinates. But the gem expects it in this format `[latitude, longitude]`. The result class returned `[longitude, latitude]`. So if you'd use this with the `geocode` method you would get the coordinates saved in the wrong attribute. Therefor I reversed the array. 

Confirmation that the coordinates in the fixture and test are now correct (or close enough ;)):

https://www.coordinatenbepalen.nl/coordinates/144787-nieuwezijds-voorburgwal-147-1012rj-amsterdam

Screenshot 2020-10-26 at 16 49 37

### DIFF
--- a/lib/geocoder/results/nationaal_georegister_nl.rb
+++ b/lib/geocoder/results/nationaal_georegister_nl.rb
@@ -8,7 +8,7 @@ module Geocoder::Result
     end
 
     def coordinates
-      @data['centroide_ll'][6..-2].split(' ').map(&:to_f)
+      @data['centroide_ll'][6..-2].split(' ').map(&:to_f).reverse
     end
 
     def formatted_address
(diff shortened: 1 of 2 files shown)


alexreisner:
Yikes. That is a serious bug. Thanks for the fix!

alexreisner:
However, the assertions you added are currently failing. Could you check the values?

alexreisner:
Perfect, thanks!

Marthyn:
No problem :) thanks for the quick merge!

PR #1474, opened 2020-11-02
Allow single stubs to be removed in test mode

jmmastey:
Hope this is a welcome possibility. I've got a test suite where folks want more fine-grained cleanup in their tests:

```ruby
  before(:all) do
    Geocoder::Lookup::Test.add_stub(
      "New York City, NY", [
        {  'coordinates'  => [40.7143528, -74.0059731] }
      ]
    )
  end

  after(:all) do
    Geocoder::Lookup::Test.delete_stub('New York City, NY')
  end
```

We didn't want to `reset` because the default stub doesn't need to change across the suite.

I took a stab at a name (`delete`, versus the word `reset` you use elsewhere, or maybe a more generic `remove`). Happy to take feedback. The unit test is also a _little_ indirect, but I think it covers the case?

### DIFF
--- a/README.md
+++ b/README.md
@@ -532,10 +532,14 @@ With the above stub defined, any query for "New York, NY" will return the result
       ]
     )
 
+You may also delete a single stub, or reset all stubs _including the default stub_:
+
+    Geocoder::Lookup::Test.delete_stub('New York, NY')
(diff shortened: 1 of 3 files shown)


alexreisner:
Yes, I like this. Thanks for the test, the docs, and the clean code!

PR #1476, opened 2020-11-14
Default cache example should use Rails.cache

richardonrails:
I'm fairly new to Rails (so let me know if I'm wrong here) but I would think that using Rails.cache makes more sense as a default, since that will allow people to use all of the other existing `cache_store` configuration, per-environment settings, etc. automatically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,7 +214,7 @@ Some common options are:
       units: :km,
 
       # caching (see [below](#caching) for details):
-      cache: Redis.new,
+      cache: Rails.cache,
       cache_prefix: "..."
 
     )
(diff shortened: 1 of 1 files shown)


alexreisner:
Similar to my note on #1477, we need to support non-Rails use, so I'm hesitant to have the README show `Rails.cache` in all config examples. Also, `Rails.cache` doesn't point to a persistent cache object in every Rails app, so it could be misleading. What if we just appended `# or Rails.cache` to each line of example code where `Redis.new` is currently shown?

richardonrails:
I think Rails.cache (regardless of persistent or not) is great for this, and then the cache can be configured in Rails.
However I understand your point regarding non-Rails. I think it'd be helpful to just have a README section about how to best integrate into Rails.

PR #1478, opened 2020-11-15
Fix country_code for Mapbox

richardonrails:
Previously Mapbox incorrectly returned the full country name when using `country_code`
Something like this should work, at least in the happy path


alexreisner:
Looks like Mapbox has added country abbreviations! Thanks for adding this. Please take a look at the test failure and I'll be happy to merge.

alexreisner:
Closing, as this was fixed by #1501.

PR #1479, opened 2020-11-21
Add coordinates to ipregistry result

richardonrails:


alexreisner:
The `coordinates` attribute is [defined in the base `Result` class](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/results/base.rb#L42), so this shouldn't be necessary. Am I missing something?

richardonrails:
Don't know why, but it doesn't work.

```rb
irb(main):047:0> result = Geocoder.search("8.8.8.8").first
=> #<Geocoder::Result::Ipregistry:0x00007fa0b76040a0 @data={..............}, @cache_hit=true>
irb(main):048:0> r.coordinates
=> [0.0, 0.0]
irb(main):049:0> [r.latitude, r.longitude]
=> [30.11718, -81.41197]
```

alexreisner:
Hmm, you're right. Looks like there are actually a couple of issues with the IPRegistry result class. Could you take a look at the test failure? Then I'll merge. Thanks!

richardonrails:
Closing since it seems like this is the wrong solution. When I opened it I didn't realize `coordinates` was defined on the base class, and I have no idea why that's not working.

alexreisner:
Sorry, I was confused (responded in a hurry, distracted, etc). The `latitude` and `longitude` methods are actually the ones implemented by the base class. The `coordinates` method does need to be implemented. My apologies. I just committed [the fix](https://github.com/alexreisner/geocoder/commit/f9af50bf48b4947c413a015519d68fc55c2be24b).

richardonrails:
Ah okay. Wonderful, thank you.

PR #1480, opened 2020-12-07
Improve ban_data_gouv_fr result

Intrepidd:
We have `region_name`, but unfortunately we don't have `region_code`. When storing regions in a database I feel like it's much better to store a technical code rather than a name, it can be error prone especially because of different ways to write it (french is weird).

Unfortunately the ban_data_gouv API does not provide with the region code, only its name.

I used the [region API](https://geo.api.gouv.fr/decoupage-administratif/regions) to get a hard-coded list of regions and their code and use it as lookup.

Those region names and code should not change so often so I figure this is an OK way to solve the problem, I don't think making an API call to get the region code would be advised, maybe if hardcoding the codes is a no-go, we could memoize or cache the API result so we only make one extra API call.

WDYT ?

### DIFF
--- a/lib/geocoder/results/ban_data_gouv_fr.rb
+++ b/lib/geocoder/results/ban_data_gouv_fr.rb
@@ -4,6 +4,27 @@ require 'geocoder/results/base'
 module Geocoder::Result
   class BanDataGouvFr < Base
 
+    STATE_CODE_MAPPINGS = {
+      "Guadeloupe" => "01",
+      "Martinique" => "02",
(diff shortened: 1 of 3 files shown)


alexreisner:
Thanks. This solution seems great to me. Merging!

Intrepidd:
Thanks ! I'd love a release, would you like some help in writing the changelog and such ?

Intrepidd:
Hi @alexreisner , gentle nudge for a new release, as stated, I can prepare some stuff if needed

PR #1492, opened 2021-03-05
Sort API list in alphabetical order

randoum:
It's much practical (and a bit ethical) to have the list alphabetically ordered.
Thanks


alexreisner:
Haha! Yep, I agree with that. Thanks.

PR #1495, opened 2021-03-11
Fix link to CONTRIBUTING.md in README

orangejulius:
Hi folks! 👋 

The documentation section on contributing was moved to a separate file in https://github.com/alexreisner/geocoder/commit/a81e028cf21d1fd31f3990ed22c466962f138a21, but the link in the main README still used anchors to the README itself for both the 'Contributing' and 'Reporting Issues' links.

This updates those links so they point to the contributing file on GitHub. Even though it's now a small document, both links point to the heading for the relevant section using anchors.


alexreisner:
Good catch! Thanks!

PR #1498, opened 2021-04-08
Add support for the new OS Data Hub to uk_ordnance_survey_names lookup

cpjmcquillan:
The Ordnance Survey API have launched [OS Data Hub](https://osdatahub.os.uk/). As part of this, users are being pushed to migrate their use to the Data Hub **before 31 May,** which uses a new API endpoint. 

This PR allows the gem to support the new endpoint alongside the existing one. I'm not sure what will happen to the existing endpoint come 31 May, but for now this will allow those who use this gem and wish to migrate to do just that.

They say that the functionality of their API hasn't changed at all. (see screenshot)

image

### DIFF
--- a/lib/geocoder/lookups/uk_ordnance_survey_names.rb
+++ b/lib/geocoder/lookups/uk_ordnance_survey_names.rb
@@ -13,7 +13,7 @@ module Geocoder::Lookup
     end
 
     def base_query_url(query)
-      "#{protocol}://api.ordnancesurvey.co.uk/opennames/v1/find?"
+      "#{protocol}://api.os.uk/search/names/v1/find?"
     end
 
     def required_api_key_parts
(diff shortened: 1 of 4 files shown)


alexreisner:
Thanks very much for this. Good PR. I think you've actually done more work than necessary. Usually when APIs change, we simply update the code to reflect the most current/future-proof version. There's usually no need to continue supporting the old URL (or syntax). In this case, it seems like anyone using the old endpoint should be automatically switched to the new one. The `:service` configuration and the new tests could be confusing in the future, once we've forgotten about the old endpoint. Updating the string returned by `base_query_url` and linking to the documentation in the Git commit should be sufficient. How do you feel about changing this to a one-line PR?

cpjmcquillan:
> Thanks very much for this. Good PR. I think you've actually done more work than necessary. Usually when APIs change, we simply update the code to reflect the most current/future-proof version. There's usually no need to continue supporting the old URL (or syntax). In this case, it seems like anyone using the old endpoint should be automatically switched to the new one. The `:service` configuration and the new tests could be confusing in the future, once we've forgotten about the old endpoint. Updating the string returned by `base_query_url` and linking to the documentation in the Git commit should be sufficient. How do you feel about changing this to a one-line PR?

Ok brilliant, thanks. 

I wasn't sure whether or not it was best to support the legacy endpoint for users until a certain point - but this is far more straightforward!

Let me know if you need anything else.

alexreisner:
Perfect, thanks!

PR #1500, opened 2021-04-30
Update Docs for Google Places API services

quorauk:
- Updates broken links
- Add new optional params to the Google Places Search Service


alexreisner:
Thank you for this!

PR #1501, opened 2021-05-06
Parse country_code and state_code on Mapbox lookups

trevorrjohn:


alexreisner:
Thanks for this. It looks basically good and merge-able but please have a look at the failing tests.

alexreisner:
It seems the Mapbox sample result that's being used by the tests is very old. I just updated it, which should make this a little easier to fix. Try pulling the latest and see if things don't work better.

trevorrjohn:
I'm having issues running it locally so that's why I'm pushing so much.
bson isn't installing on my Mac...

PR #1502, opened 2021-05-12
feat: add configuration option for maxminddb local_ip_alias

coder2000:
Adds a configuration option to use the local_ip_alias attribute of the maxminddb library to use a local IP address to geolocate during testing. I don't use the Hive::GeoIP2 gem to test it with but the configuration option will only work if the class responds to the attribute so it shouldn't fail if it doesn't exist.


alexreisner:
Looks good to me. Thanks!

PR #1508, opened 2021-06-15
Update ip2location lookup for v2 API

jamesbebbington:
Just a little PR to update the `ip2location` lookup to use [their v2 API](https://www.ip2location.com/web-service/ip2location).

### DIFF
--- a/lib/geocoder/lookups/ip2location.rb
+++ b/lib/geocoder/lookups/ip2location.rb
@@ -23,11 +23,11 @@ module Geocoder::Lookup
     end
 
     def query_url_params(query)
-      {
+      super.merge(
         key: configuration.api_key,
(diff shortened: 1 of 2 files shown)


alexreisner:
Thanks for this! It looks good except I think the way you're merging params in `query_url_params`, `super` will overwrite the local config, which is the opposite of what we want. Can you take a look at that?

jamesbebbington:
There you go, let me know if you want that commit squashed into previous.

alexreisner:
Oof, thanks for pointing out Baidu. That's definitely wrong. And thanks for fixing! Will merge shortly.

jamesbebbington:
Brilliant! Thanks Alex.

PR #1509, opened 2021-06-16
IPQualityScore IP lookup

jamesbebbington:
This PR adds support for IP lookups via [IPQualityScore](https://www.ipqualityscore.com), a full-featured service with comprehensive IP reputation data.


alexreisner:
This looks great! There are some failing tests, though. Could you take a look at them? For attributes not supplied by the API (eg, `state_code`), the standard behavior is to return the most relevant piece of data (eg, state full name). This behavior probably isn't ideal, but we need to maintain it for consistent behavior across all APIs.

jamesbebbington:
Thanks! Sure, no problem, though can you be more specific about which test failures you're seeing? I've only been testing `test/unit/lookups/ipqualityscore_test.rb` (which all passes for me) as I've not been able to get the full test suite to run, and when I exclude the test files that break the full suite for me, I see 1 failure and 14 errors across many unrelated tests files.

Relatedly, is the Travis CI integration broken? That would help a lot, if it could be coaxed back to life?

jamesbebbington:
Scratch that, I've managed to get the full suite to run now and can see a bunch of grumbles about a missing `ipqualityscore_madison_square_garden` fixture file. Will fix.

jamesbebbington:
There you go, full suite now green.

alexreisner:
Looks good. Thanks!

PR #1512, opened 2021-06-30
Amazon Location Service provider

### LINKED ISSUE #1511: Support for Amazon Location Service
AWS recently launched [Amazon Location Service](https://aws.amazon.com/location/) which provides geocoding and reverse geocoding.

I would be happy to try and implement the Amazon client for Geocoder. Are you interested in adding support for Amazon Location Service?

mplewis:
Closes #1511.

This is an initial and incomplete attempt to incorporate Amazon Location Service into Geocoder. I am opening this PR to start a conversation about how I can help move this forward.

This implementation uses the AWS SDK to provide access to the API, rather than the traditional REST API seen throughout this repo. I made this choice because AWS developers have a lot of ways to configure apps that use AWS. They tend to use environment variables like `AWS_ACCESS_KEY_ID`, but they can also use the local system configuration or implicit roles provided by EC2 at execution time.

I thought it was appropriate to ask developers to install part of the AWS SDK so that we can allow Amazon to take care of authorization and signing. I included a note in the readme about this.

This PR lacks tests as I wasn't sure how to run the test suite and Travis doesn't seem to be working at this time. I would appreciate guidance on how to run these tests so that I can write tests that match the AWS SDK and its behavior.

### DIFF
--- a/README_API_GUIDE.md
+++ b/README_API_GUIDE.md
@@ -14,7 +14,7 @@ Table of Contents
 Global Street Address Lookups
 -----------------------------
 
-### Amazon Location Service (`:amazon`)
+### Amazon Location Service (`:amazon_location_service`)
 
 * **API key**: required
 * **Key signup**: https://console.aws.amazon.com/location
--- a/README_API_GUIDE.md
+++ b/README_API_GUIDE.md
@@ -22,14 +22,37 @@ Global Street Address Lookups
 * **Region**: world
 * **SSL support**: yes, required
 * **Languages**: en
+* **Required params**:
+  * `:index_name` - the name of the place index resource you want to use for the search
 * **Extra params**:
   * `:max_results` - return at most this many results
-  * `:bias_position` - bias the results toward a given point, defined as `[latitude, longitude]`
-  * `:filter_b_box` - a bounding box that you specify to filter your results to coordinates within the box's boundaries, defined as `[longitude_sw, latitude_sw, longitude_ne, latitude_ne]`
-  * `:filter_countries` - an array of countries you want to geocode within, named by [ISO 3166 country codes](https://www.iso.org/iso-3166-country-codes.html), e.g. `['DEU', 'FRA']`
+* **Extra params** when geocoding (not reverse geocoding):
+    * `:bias_position` - bias the results toward a given point, defined as `[latitude, longitude]`
+    * `:filter_b_box` - a bounding box that you specify to filter your results to coordinates within the box's boundaries, defined as `[longitude_sw, latitude_sw, longitude_ne, latitude_ne]`
+    * `:filter_countries` - an array of countries you want to geocode within, named by [ISO 3166 country codes](https://www.iso.org/iso-3166-country-codes.html), e.g. `['DEU', 'FRA']`
 * **Documentation**: https://docs.aws.amazon.com/location
 * **Terms of Service**: https://aws.amazon.com/service-terms
-* **Notes**: You must install either the `aws-sdk` or `aws-sdk-locationservice` gems, version 1.4.0 or greater. TODO: AWS auth
+* **Notes**:
+  * You must install either the `aws-sdk` or `aws-sdk-locationservice` gem, version 1.4.0 or greater.
+  * You can set a default index name for all queries in the Geocoder configuration:
+    ```rb
+      Geocoder.configure(
+        lookup: :amazon,
+        index_name: 'YOUR_INDEX_NAME_GOES_HERE',
+      )
+    ```
+  * You can provide credentials to the AWS SDK in multiple ways:
+    * Directly via the `api_key` parameter in the Geocoder configuration:
+      ```rb
+        Geocoder.configure(
+          lookup: :amazon,
+          api_key: {
+            access_key_id: 'YOUR_AWS_ACCESS_KEY_ID_GOES_HERE',
+            secret_access_key: 'YOUR_AWS_SECRET_ACCESS_KEY_GOES_HERE',
+          },
+        )
+      ```
+    * Via environment variables and other external methods. See **Setting AWS Credentials** in the [AWS SDK for Ruby Developer Guide](https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/setup-config.html).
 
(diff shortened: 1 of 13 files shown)


mplewis:
@alexreisner, were you able to take a look at this PR? I'd love some feedback whenever you have time.

alexreisner:
Thanks for the good work on this, and sorry for the delay. I agree that an external gem should handle AWS's unusual authentication. Some other thoughts:

* I think it should be called `:amazon_location_service`, just in case Amazon adds another service that Geocoder wants to connect to.
* We should remove the three lines at the top of the lookup class that require external gems (pry, etc).
* In the `Geocoder::Result` subclass, we should _not_ define `latitude` and `longitude` methods (only `coordinates` is needed).
* We need some basic tests, or at least some fixtures with sample responses.

Other than that, it looks great!

mplewis:
Thanks. I'm happy to write some tests but I'm having trouble running the test suite in this project. Could you share any instructions on how to run tests for Geocoder?

alexreisner:
It should be as simple as:

    bundle install
    bundle exec rake test

What errors are you getting?

mplewis:
I currently get the following error after attempting `bundle install`:

```
Bundler could not find compatible versions for gem "bson":
  In Gemfile:
    bson_ext arm64-darwin-20 was resolved to 1.12.5, which depends on
      bson (~> 1.12.5) arm64-darwin-20

    mongoid arm64-darwin-20 was resolved to 1.0.6, which depends on
mongo (>= 0.18.2) arm64-darwin-20 was resolved to 2.1.2, which depends
on
        bson (~> 3.0)
```

alexreisner:
Hmm, I'm using the following versions:

    bson 4.8.2
    bson_ext 1.5.1
    mongoid 7.1.0

Can you try locking to those and see if they install?

mplewis:
Thanks for taking a look. I tried the versions you gave me and it resulted in further dependency conflicts.

I think I know what the problem is here; I'm on an ARM MacBook and the versions of the packages I need aren't available for this OS. Hopefully some `bundle upgrade`s can clear the issue up now that the ARM machines have been out for a while?

I will try and finish out tests for this provider in a VM and see if I can `bundle install` there.

mplewis:
@alexreisner, I was able to install dependencies on my ARM machine after using Ruby 2.7.2 instead of 3.x. I still had some trouble with MySQL so I had to stub it out – just a heads up.

I wrote some tests and a fixture that match live Amazon Location Service data. Please take a look and let me know what you think.

mplewis:
@alexreisner, friendly ping – anything I can do to help merge this?

alexreisner:
Sorry for the delay on this. Definitely still interested in merging. Appreciate your patience. Just a few more things:

The `index_name` config param is only used by this lookup, so there's actually no need to add it to the array of params in the `Configuration` class. Recommended usage should be to put it in an Amazon-specific part of the config, eg:

    Geocoder.configure(
        lookup: :amazon_location_service,
        amazon_location_service: {
          index_name: ...,
          api_key: ...
        }
      )

The `global_index_name` method in the lookup is also unnecessary, you can access it via `configuration.index_name` within the lookup.

I'm also getting test problems: 6 failures, 5 errors. Could you look into that?

mplewis:
@alexreisner, thanks for the comments. I'll push a commit to fix the issues you described.

Unfortunately I cannot run the test suite locally because I do not know how to set up all the dependencies on my system. Would you mind posting the test failure logs so I can work to fix the ones you see?

alexreisner:
No problem. I've pasted one of the errors below, and I just fixed the Travis integration so you should see the test suite being run after every subsequent push. Let me know if that doesn't happen.

    /geocoder/test/unit/cache_test.rb:19:in `test_second_occurrence_of_request_is_cache_hit'
    /geocoder/test/unit/cache_test.rb:19:in `each'
         20:       next if l == :maxmind_local || l == :geoip2 # local, does not use cache
         21:       Geocoder.configure(:lookup => l)
         22:       set_api_key!(l)
      => 23:       results = Geocoder.search("Madison Square Garden")
         24:       assert !results.first.cache_hit,
         25:         "Lookup #{l} returned erroneously cached result."
         26:       results = Geocoder.search("Madison Square Garden")
    /geocoder/test/unit/cache_test.rb:23:in `block in test_second_occurrence_of_request_is_cache_hit'
    /geocoder/lib/geocoder.rb:22:in `search'
    /geocoder/lib/geocoder/query.rb:11:in `execute'
    /geocoder/lib/geocoder/lookups/base.rb:46:in `search'
    /geocoder/lib/geocoder/lookups/amazon_location_service.rb:14:in `results'
    Error: test_second_occurrence_of_request_is_cache_hit(CacheTest): NoMethodError: undefined method `results' for nil:NilClass

mplewis:
@alexreisner, I've fixed the breaking tests, so now the TravisCI test suite is all green.

One change you may be concerned about is where I change `all_services_except_test` out for `all_services_with_http_requests` in the `error_handling_test`. I excluded AWS ALS from this spec because it doesn't seem to apply, but please feel free to change this implementation if you prefer it a different way.

I think that is everything so far. Thanks for the feedback – let me know if you have any more.

alexreisner:
Great, thanks. The `all_services_with_http_requests` solution for testing looks reasonable to me. Just a couple more minor issues and then I think we're ready to merge:

1. Maybe I'm missing something, but what's gained by the assignment `@configuration = configuration`? It seems like it makes the Lookup object kind of brittle: you have to call `results` (where the assignment is made) before calling other methods like `client`, which depend on the assignment of `@configuration`. They're private methods, so in practice this might not matter, but still, it seems confusing to me.
2. Lookups should not define the `latitude` and `longitude` methods--these are defined in the base lookup. Subclass lookups should only define `coordinates`, which should return both values as an array.

mplewis:
@alexreisner Thanks for your patience. Both of those changes sound good to me. I can't remember why I tried to memoize `@configuration`, so I pulled it out. Tests are still passing as before.

Let me know if there is anything else you find that blocks merging.

alexreisner:
This looks great. Thanks for _your_ patience as well. Merging shortly!

PR #1513, opened 2021-07-07
Accept IPAddr when initializing an IpAddress

szechyjs:
If using `geocoded_by` with an ActiveRecord column with an `inet` type the value is automatically converted to an `IPAddr`. This then prevents geocoder from recognizing it as an IP, as it expects a string.

This updates the `IpAddress` initializer to accept  and convert an `IPAddr`.


alexreisner:
Excellent, thanks!

PR #1514, opened 2021-07-16
Add support for fields param to google_places_details lookup

ericds:
Hi @alexreisner, 

I modeled support for a fields param after the fields param in the [google_places_search](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookups/google_places_search.rb) lookup. 

This adds support for a fields param without changing existing behavior if you don't pass a fields param.

Please let me know if you're open to adding support for a fields param to this lookup and if you'd like to see any changes. If you're open to the change, I could also update the readme as a part of the PR.

Thanks!

### DIFF
--- a/lib/geocoder/lookups/google_places_details.rb
+++ b/lib/geocoder/lookups/google_places_details.rb
@@ -36,8 +36,30 @@ module Geocoder
       def query_url_google_params(query)
         {
           placeid: query.text,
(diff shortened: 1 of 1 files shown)


alexreisner:
Thanks for this, and sorry I somehow missed it. Functionality was added by #1572.

PR #1522, opened 2021-08-25
IPQualityScore unsuccessful response caching fix

jamesbebbington:
Hey @alexreisner, another small IPQualityScore fix for you; this one ensures that only successful responses are cached, when caching is enabled.


alexreisner:
Excellent!

PR #1523, opened 2021-09-01
Remove duplicate Geocoder.configure

ALacker:
I made a copy of these tests to start which is how I introduced the typo here: https://github.com/alexreisner/geocoder/pull/1521#discussion_r695942405

So I figured I'd remove it from this file so no one else makes the same copy-pasta mistake


alexreisner:
Haha! Amazing, thank you.

PR #1532, opened 2021-10-27
Open for extensions + redis expiration

shqear93:
Applying open for extensions principle
and Adding support for redis expiration (ttl)

usage:
```ruby
Geocoder.configure(
  lookup: :google,
  api_key: ENV['GMAPS_PLAECS_KEY'].try(:freeze),
  cache: Redis.current,
  cache_prefix: "GeocoderCache:",
  expire: 2.days
)
```

### DIFF
--- a/lib/generators/geocoder/config/templates/initializer.rb
+++ b/lib/generators/geocoder/config/templates/initializer.rb
@@ -10,7 +10,7 @@ Geocoder.configure(
   # api_key: nil,               # API key for geocoding service
   # cache: nil,                 # cache object (must respond to #[], #[]=, and #del)
   # cache_prefix: 'geocoder:',  # prefix (string) to use for all cache keys
-  # expire: 1.day,              # expiration time (ttl)
+  # cache_expiration: 1.day,    # expiration time (ttl)
 
   # Exceptions that should not be rescued by default
   # (if you want to implement custom error handling);
--- a/lib/generators/geocoder/config/templates/initializer.rb
+++ b/lib/generators/geocoder/config/templates/initializer.rb
@@ -9,7 +9,7 @@ Geocoder.configure(
   # https_proxy: nil,           # HTTPS proxy server (user:pass@host:port)
   # api_key: nil,               # API key for geocoding service
   # cache: nil,                 # cache object (must respond to #[], #[]=, and #del)
-  # cache_prefix: 'geocoder:',  # - DEPRECATED - prefix (string) to use for all cache keys
+  # cache_prefix: 'geocoder:',  # - DEPRECATED - prefix (string) to use for all cache keys, set false to disable
 
   # Exceptions that should not be rescued by default
   # (if you want to implement custom error handling);
--- a/lib/generators/geocoder/config/templates/initializer.rb
+++ b/lib/generators/geocoder/config/templates/initializer.rb
@@ -10,6 +10,7 @@ Geocoder.configure(
   # api_key: nil,               # API key for geocoding service
   # cache: nil,                 # cache object (must respond to #[], #[]=, and #del)
   # cache_prefix: 'geocoder:',  # prefix (string) to use for all cache keys
+  # expire: 1.day,              # expiration time (ttl)
 
   # Exceptions that should not be rescued by default
   # (if you want to implement custom error handling);
--- a/lib/generators/geocoder/config/templates/initializer.rb
+++ b/lib/generators/geocoder/config/templates/initializer.rb
@@ -9,8 +9,7 @@ Geocoder.configure(
   # https_proxy: nil,           # HTTPS proxy server (user:pass@host:port)
   # api_key: nil,               # API key for geocoding service
   # cache: nil,                 # cache object (must respond to #[], #[]=, and #del)
-  # cache_prefix: 'geocoder:',  # prefix (string) to use for all cache keys
-  # cache_expiration: 1.day,    # expiration time (ttl)
+  # cache_prefix: 'geocoder:',  # - DEPRECATED - prefix (string) to use for all cache keys
 
   # Exceptions that should not be rescued by default
   # (if you want to implement custom error handling);
@@ -20,4 +19,13 @@ Geocoder.configure(
   # Calculation options
   # units: :mi,                 # :km for kilometers or :mi for miles
   # distances: :linear          # :spherical or :linear
+
+  # Cache service type-specific configurations
+  # cache_options: {}
+
+  # Redis cache configurations (for cache_options):
+  #     {
+  #       expiration: 2.days          # redis ttl for all cache
+  #       prefix: 'geocoder:'         # prefix to add to redis keys, if used cache_prefix will be ignored
+  #     }
 )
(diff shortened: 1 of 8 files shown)


shqear93:
@alexreisner please let me know what do you think about this 🙏

shqear93:
currently I'm also using this to save to databse instead of redis
```ruby
Geocoder.configure(
  lookup: :google,
  api_key: ENV['GMAPS_PLAECS_KEY'].try(:freeze),
  cache: GeocoderDatum,
  cache_prefix: ""
)
```

```ruby
class GeocoderDatum < ApplicationRecord
  include GeocoderDatum::ServiceConcern

  serialize :data, Serializers::IndifferentHash
end 

module GeocoderDatum::ServiceConcern
  extend ActiveSupport::Concern

  class_methods do
    def [](url)
      find_by_url(url)&.data_before_type_cast
    end

    def []=(url, value)
      where(url: url).first_or_create(data: value).update(data: value)
    end

    def keys
      pluck(:url)
    end
  end
end 

```

lib/geocoder/lookups/base.rb
```diff
@@ -84,7 +84,7 @@ def query_url(query)
       #
       def cache
         if @cache.nil? and store = configuration.cache
-          @cache = Cache.new(store, configuration.cache_prefix)
+          @cache = Cache.new(store, configuration.except(:cache))
```
alexreisner, line 87: I think we should be a little more careful here and only pass the relevant options. There are only two, right?

lib/geocoder/cache_stores/other.rb
```diff
@@ -0,0 +1,33 @@
+module Geocoder::CacheStore
+  class Other < Base
```
alexreisner, line 2: Can we call this `Generic` instead of `Other`?

alexreisner (changes requested):
Again, I really like this. Thanks for the work you've put into it. Just a few comments, mostly about naming, for clarity and consistency. Happy to discuss further.

lib/geocoder/cache.rb
```diff
@@ -1,23 +1,18 @@
+Dir["#{__dir__}/cache_stores/*.rb"].each {|file| require file }
+
 module Geocoder
   class Cache
 
-    def initialize(store, prefix)
-      @store = store
-      @prefix = prefix
+    def initialize(store, configs)
+      @class = ("Geocoder::CacheStore::#{store.class}".constantize rescue Geocoder::CacheStore::Other)
```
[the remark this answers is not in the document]
shqear93: @alexreisner I'll change to `Object.const_get("Geocoder::CacheStore::#{store.class}")`

lib/geocoder/cache_stores/other.rb
shqear93: @alexreisner yup that makes more sense, I'll update it

lib/geocoder/lookups/base.rb
shqear93: @alexreisner I thought about passing all options just in case they're been used by different services, so if influxdb support is going to be added later -as example- it might have it's own configs which is might be different than redis options

alexreisner: That is definitely true. However, I still think it's better to pass the arguments explicitly, and add new ones as new caching services require them. Otherwise we're passing options like `:use_https` to the cache service, which is potentially confusing because it sounds like a valid option (connect to Redis via HTTPS?) but certainly not what the user intended (connect to geocoding API via HTTPS). The same goes for `:logger` and `:host`...

Maybe a better solution, to allow for expansion in the future, is to add a `:cache_options` config param with a hash value instead of `:cache_expiration`. So config would look something like this:

    Geocoder.configure(
      cache: Redis.current,
      cache_options: {
        expiration: 2.days,
        prefix: 'geocoder:',
        other_option: ...
      }
    )

and the whole `cache_options` hash would be passed to the cache service. For backward compatibility, we'd need to continue supporting the `:cache_prefix` option at the root level... What do you think of this?

shqear93: @alexreisner I think you're totally right, I also agree on your suggestion, I'll implement it

shqear93: @alexreisner I think you're totally right, I also agree on your suggestion, but I think going with that suggestion will break backward compatibility, so people who were using `cache_prefix` will have a toubles, so I think we can leave `cache_prefix` outside for now, and also I'll copy it inside cache_options if not provided just as a compatibility trick, what do you think?
```ruby
Geocoder.configure(
  lookup: :google,
  api_key: ENV['GMAPS_PLAECS_KEY'].try(:freeze),
  cache: Redis.current,
  cache_prefix: "GeocoderCache:"
)```

Comment (shqear93) [reply lib/geocoder/lookups/base.rb]: @alexreisner what do you thinks about https://github.com/alexreisner/geocoder/pull/1532/commits/72a6e379fcd449835004d4854f6df30589c94ace ?

File: lib/geocoder/lookups/base.rb [hunk 4]
```diff
@@ -83,8 +83,13 @@ def query_url(query)
       # The working Cache object.
       #
       def cache
-        if @cache.nil? and store = configuration.cache
-          @cache = Cache.new(store, configuration.cache_prefix)
+        if @cache.nil? && (store = configuration.cache)
+          cache_options = configuration.cache_options
+          if configuration.cache_prefix.present?
```
alexreisner, line 88: This functionality is good, but the `present?` method is provided by Rails, so we can't use it here.

.gitignore
```diff
@@ -6,3 +6,4 @@ rdoc/*
 Gemfile.lock
 api_keys.yml
 test/geocoder_test.sqlite
+.idea
```
alexreisner, line 9: This shouldn't be here because it's not Geocoder-specific. Please add to your `~/.gitignore_global` instead.

lib/generators/geocoder/config/templates/initializer.rb
```diff
@@ -10,6 +10,7 @@
   # api_key: nil,               # API key for geocoding service
   # cache: nil,                 # cache object (must respond to #[], #[]=, and #del)
   # cache_prefix: 'geocoder:',  # prefix (string) to use for all cache keys
+  # cache_expiration: 1.day,    # expiration time (ttl)
```
alexreisner, line 13: We should probably create only `cache_options` here (no _prefix or _expiration).

alexreisner (changes requested):
Looks good! Thanks again. Just a few comments.

lib/generators/geocoder/config/templates/initializer.rb
shqear93: @alexreisner I see I forgot to remove `cache_expiration` from docs, for `cache_prefix` it's for backward compatibility, please check my comment here https://github.com/alexreisner/geocoder/pull/1532#discussion_r745106020 and let me know

shqear93:
@alexreisner let me know what do you think about my last updates ✌️

lib/geocoder/lookups/base.rb
shqear93: @alexreisner let me know your feedback about https://github.com/alexreisner/geocoder/pull/1532/commits/5426ff13ad7d4138a21c300c19b848dec2b52344

alexreisner:
Looks great! Thanks for the good work on this.

PR #1533, opened 2021-11-04
Update README_API_GUIDE.md

titanium-cranium:
ipstack has recently changed their pricing plans, updated info

Oddly, I've seen both 5000 requests/mo on their pricing page, and at other times 100.  

The suggested change alters their quota to 100 based on recent experience.

Actual limit 100 requests per month, my free account data
Screen Shot 2021-11-04 at 2 31 57 pm


alexreisner:
Very interesting. Thanks for this!

PR #1534, opened 2021-11-08
add https support for pointpin

jurgens:
it seems that when people are using pointpin service via geocoder it uses http protocol while we're supporting https for quite a while

### DIFF
--- a/lib/geocoder/lookups/pointpin.rb
+++ b/lib/geocoder/lookups/pointpin.rb
@@ -52,7 +52,7 @@ module Geocoder::Lookup
     end
 
     def supported_protocols
-      [:https]
+      [:http, :https]
     end
 
     # TODO: replace this hash with what's actually returned by Pointpin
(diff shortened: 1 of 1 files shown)


alexreisner:
Thanks for this. My concern here is that Pointpin does still support HTTP (there are even [examples in the docs](https://pointp.in/docs/get-started)) and this PR makes it impossible for Geocoder to use it. My preference is to let users choose what protocol they want to use via Geocoder config. Are there plans to phase out non-HTTPS access? Is there another issue with the HTTP requests?

jurgens:
@alexreisner thanks for quick response
I've re-enabled http 

I will update documentation though to mention https only

alexreisner:
The method is now identical to [the one inherited from the Base lookup](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookups/base.rb#L104) so I'm going to close this. But do please help us keep Geocoder up to date as Pointpin evolves!

jurgens:
ah, I didn't see this method
but now I don't understand why one of our clients is having troubles connecting via http if https is enabled and is preferable in geocoder gem

alexreisner:
Hmm, that is strange. The way to do it is to set `use_https` when configuring Geocoder, eg:

    Geocoder.configure(use_https: true)

PR #1539, opened 2021-12-05
add support for host parameter to ipdata_co lookup

gokce:
ipdata introduced an EU endpoint for API calls. https://docs.ipdata.co/
This pull request adds support for passing host parameter to ipdata_co lookup.


alexreisner:
Excellent, thanks.

PR #1551, opened 2021-12-10
ArgumentError: unexpected value at params[:lookup]

defkode:
aws-sdk-locationservice raises and error when :lookup parameter is given:

```ArgumentError: unexpected value at params[:lookup]```

This Pr fixes this and improves readability


alexreisner:
Thanks for this. It looks good except for one thing: the `Hash#except` method was added in Ruby 3.0, and we need to support 2.x.

alexreisner:
Thanks!

jdongelmans:
Any timeframe for a release of this? Just encountered this error while working with the AWS lookup 🙂 
Thanks for the great gem by the way!

alexreisner:
@jdongelmans Just released. Happy new year!

PR #1561, opened 2022-01-17
Move cache_prefix config option to cache_options

xtr3me:
When using the Gem on version 1.7.2, i've noticed that when using the cache option it will by default spit out a deprecation warning to not use the configuration option: `cache_prefix` anymore.

```
[Geocoder] cache_prefix is deprecated, please change to cache_options.prefix instead
```

This PR moves the default value for `cache_prefix` to its new desired location so that the deprecation warning will be removed when a developer is not explicitly setting the incorrect option.

See my other PR for a documentation change in the README.md to use the correct location


alexreisner:
More good work! Much appreciated.

Issue #1556, opened 2021-12-27
Nominatim methods throw NoMethodError when @data['address'] is nil


duarme:
### Expected behavior

When calling `result.city`, if there is no `city`, [Nominatim#city](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/results/nominatim.rb#L26) should return nil.

### Actual behavior

When `@data['address']` is nil, it throws a **NoMethodError: undefined method `key?' for nil:NilClass**:

![CleanShot 2021-12-27 at 11 30 41](https://user-images.githubusercontent.com/794154/147467731-f78935a9-c41f-4d56-99ce-ca2de3449ea2.png)

### Steps to reproduce

Try to geocode any result that doesn't have `@data['address']` and call `Nominatim#city` on it (or any other similar method).

### Environment info

* Geocoder version: 1.6.7
* Rails version:  5.2.4.3
* Database (if applicable): postgres
* Lookup (if applicable): locationiq

### Proposed solution

How about switching 

```ruby
[...] if @data['address'].key?(key)
```
 
 to 
 
 ```ruby
 [...] if @data['address']&.key?(key)
 ``` 
 
 into [nominatim.rb](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/results/nominatim.rb) ?

alexreisner:
Thanks for this. Your proposed solution will work for the `city` method but there are more than a dozen others that rely on `@data['address']` being non-nil, and this would be a good time to fix all of them. Out of curiosity, do you have a sample Nominatim response where geocoding is successful but address is empty? Would be good to add a [test fixture](https://github.com/alexreisner/geocoder/tree/master/test/fixtures).

duarme:
You're welcome. Oddly enough, I don't, I couldn't replicate this because every time I retry, it works. It seems a glitch in the lookup service, in order to catch it I should add some debug line and run them in production. Anyway, I'm not sure that's what's going on. I mean, how do you know that geocoding is successful? Before the failing line I'm calling `result.latitude` and `result.longitude`, but they could be nil and I'd get `nil.to_f => 0.0`, won't I?

```ruby
      def coordinates
        [@data['latitude'].to_f, @data['longitude'].to_f]
      end

      def latitude
        coordinates[0]
      end

      def longitude
        coordinates[1]
      end
```

Let me find a Search with `latitude = 0.0`:

```
irb(main):019:0> s = Search.find(147274455)
[...]

PR #1562, opened 2022-01-17
#1556 Made the Nominatim parsing more robust

xtr3me:
To mitigate the issues reported in https://github.com/alexreisner/geocoder/issues/1556
I've moved the direct lookup of `address` within `@data` to a new private method `#address_data` that always returns a Hash if the result is nil, allowing the lookup methods to return nil instead of an error.

Also the `#coordinates` method now returns an empty array, which will be picked up by `#geocoded?` as false:

```ruby
module Geocoder
  module Store
    module Base
        ##
        # Is this object geocoded? (Does it have latitude and longitude?)
        #
        def geocoded?
          to_coordinates.compact.size == 2
        end
    end
  end
end

```

I choose the `#address_data` method name as the `#address` method is already used within other result objects as a public method.

I was unable to run the test suite on my M1, so that still needs some checking.

### DIFF
--- a/lib/geocoder/results/nominatim.rb
+++ b/lib/geocoder/results/nominatim.rb
@@ -4,7 +4,7 @@ module Geocoder::Result
   class Nominatim < Base
 
     def poi
-      return address_data[place_type] if address.key?(place_type)
+      return address_data[place_type] if address_data.key?(place_type)
       return nil
     end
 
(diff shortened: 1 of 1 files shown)


lib/geocoder/results/nominatim.rb
```diff
@@ -18,69 +18,71 @@ def address
 
     def street
       %w[road pedestrian highway].each do |key|
-        return @data['address'][key] if @data['address'].key?(key)
+        return address_data[key] if address.key?(key)
```
alexreisner, line 21: In the `if` clause, here and on line 28, it should be `address_data`.

alexreisner (changes requested):
This looks good except for the variable name identified below. Thanks @xtr3me!

xtr3me:
Thanks! that was a leftover after renaming the method, when I saw it was already used in other result classes.

PR #1564, opened 2022-02-02
Freegeoip abides by base class recommendations and supports an api key

joeyAghion:
The `Lookup::Base` class [recommends](https://github.com/alexreisner/geocoder/blob/31153d518687351b68a5d0abd5e087f47e6f4204/lib/geocoder/lookups/base.rb#L74-L76) that implementations _not_ override the `query_url` method, but instead provide `base_query_url` and `url_query_string` methods. This PR brings the Freegeoip provider inline with that recommendation.

This was necessary in order to add support for an `apikey` parameter, required by some services such as freegeoip.app.

### DIFF
--- a/lib/geocoder/lookups/freegeoip.rb
+++ b/lib/geocoder/lookups/freegeoip.rb
@@ -17,14 +17,16 @@ module Geocoder::Lookup
       end
     end
 
-    def query_url(query)
-      "#{protocol}://#{host}/json/#{query.sanitized_text}"
-    end
-
     private # ---------------------------------------------------------------
 
-    def cache_key(query)
-      query_url(query)
+    def base_query_url(query)
+      "#{protocol}://#{host}/json/#{query.sanitized_text}?"
+    end
+
+    def query_url_params(query)
+      {
+        :apikey => configuration.api_key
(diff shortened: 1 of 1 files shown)


alexreisner:
Thanks for this! I haven't looked at the FreeGeoIP situation in a while, but it appears [the README](https://github.com/alexreisner/geocoder/blob/master/README_API_GUIDE.md#freegeoip-freegeoip---discontinued) is out of date. As someone who uses this service, would you say `freegeoip.app` is a good default value for `:host`? It looks like we can use `api.ipstack.com` too, and I'm not sure which is a better choice.

joeyAghion:
I'm honestly not sure, having only learned about freegeoip.app from https://github.com/alexreisner/geocoder/pull/1382.

IPStack's API appears to be a little different, so it probably makes sense for that to be a separate provider. Maybe it [already is](https://github.com/alexreisner/geocoder/blob/master/README_API_GUIDE.md#ipstack-ipstack)?

alexreisner:
OK, thanks, I'll look into the FreeGeoIP situation some more. At the very least, it seems like the "DISCONTINUED" notice should be removed...

PR #1567, opened 2022-03-08
Only require Lookup classes if not defined to allow lookup extensions

januszm:
This will allow for extending Geocoder with additional lookups that come from the application, not shipped with the gem. 

```ruby
# in app/libs/geocoder/lookup/my_preciousss.rb
module Geocoder::Lookup
  class MyPreciousss < Google
  end
end

# config/initializers/geocoder.rb
Geocoder::Lookup.street_services << :my_preciousss
Geocoder::Lookup.street_services << :my_other

# > Geocoder.configure(my_preciousss: { api_key: 'abcdef' })
# > Geocoder.configure(my_other: { api_key: 'fedabc' })

# > Geocoder.search("Paris", lookup: :my_preciousss)
```

An example use case for this is, when we have several groups of users in the application, or tenants who use Google paid services and we want to properly separate these paid services. Since Geocoder Configuration is a `Singleton`, we cannot do this by creating an "instance" of the Geocoder with different api_key(s). As a workaround, you can define Google Lookup subclasses and define a different api_key for each of them.

The main problem I am trying to solve here is that you cannot define different api_keys for particular lookup service for different geocoder instances in one process at runtime.

### DIFF
--- a/examples/extend_lookup_services.rb
+++ b/examples/extend_lookup_services.rb
@@ -1,0 +1,22 @@
+# To extend the Geocoder with additional lookups that come from the application,
+# not shipped with the gem, define a "child" lookup in your application, based on existing one.
+# This is required because the Geocoder::Configuration is a Singleton and stores one api key per lookup.
+
+# in app/libs/geocoder/lookup/my_preciousss.rb
+module Geocoder::Lookup
+  class MyPreciousss < Google
+  end
+end
+
+# Update Geocoder's street_services on initialize:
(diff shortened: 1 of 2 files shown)


alexreisner:
@januszm this is a very nice solution! I'd only suggest one thing: can we change the name of the new method to `instantiate_lookup`?

januszm:
@alexreisner it's done!

alexreisner:
Thanks!

PR #1575, opened 2022-04-26
Update HERE lookup to use Geocoder and Search API

Pritilender:
@alexreisner Please take a look.

Docs for new API:
* [Guide on geocoding](https://developer.here.com/documentation/geocoding-search-api/dev_guide/topics/endpoint-geocode-brief.html)
* [Guide on reverse geocode](https://developer.here.com/documentation/geocoding-search-api/dev_guide/topics/endpoint-reverse-geocode-brief.html)
* [API reference](https://developer.here.com/documentation/geocoding-search-api/api-reference-swagger.html)

Some things to note regarding new API:
* It doesn't support bounding box as query param. For geocoding it supports the country code, and for reverse geocode it supports a circular area.
* For `resultType: place`, the result is missing `viewport` now (that's why I've added a Berlin fixture)

### DIFF
--- a/lib/geocoder/lookups/here.rb
+++ b/lib/geocoder/lookups/here.rb
@@ -19,48 +19,40 @@ module Geocoder::Lookup
     private # ---------------------------------------------------------------
 
     def base_query_url(query)
-      "#{protocol}://#{if query.reverse_geocode? then 'reverse.' end}geocoder.ls.hereapi.com/6.2/#{if query.reverse_geocode? then 'reverse' end}geocode.json?"
+      service = query.reverse_geocode? ? "revgeocode" : "geocode"
+
+      "#{protocol}://#{service}.search.hereapi.com/v1/#{service}?"
     end
 
     def results(query)
       return [] unless doc = fetch_data(query)
-      return [] unless doc['Response'] && doc['Response']['View']
-      if r=doc['Response']['View']
-        return [] if r.nil? || !r.is_a?(Array) || r.empty?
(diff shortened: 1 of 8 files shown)


lib/geocoder/lookups/here.rb
```diff
@@ -19,48 +19,40 @@ def supported_protocols
...
 
     def results(query)
       return [] unless doc = fetch_data(query)
-      return [] unless doc['Response'] && doc['Response']['View']
-      if r=doc['Response']['View']
-        return [] if r.nil? || !r.is_a?(Array) || r.empty?
-        return r.first['Result']
-      end
-      []
+      return [] unless doc["items"] && !doc["items"].empty?
```
januszm, line 29: `return x unless A && B` is always a bit hard to read (I even think there might be a rubocop for this), perhaps we could rewrite it with `return [] if ...` ?

alexreisner:
This looks good! Thanks for the useful links and notes.

I have one suggestion. I just tried to use the Here API with my old key, which is a 2-element array. The API returned a 500 error, without any message. For user-friendliness, maybe we should check for an array value in the `:api_key` config and, if one is detected, print a message about Here's new API, the need for a new key, etc. I think the `results` method (in the Lookup) would be a good place to do this, but I'm open to suggestions.

Pritilender:
Should be fixed now :)

alexreisner:
Perfect, thanks! I just merged this in to the `v1.8` branch. To be released soon.

PR #1577, opened 2022-04-27
Add locationbias parameter to Google Places Search

czlee:
Hi again! I needed to add the `locationbias` parameter for my work and figured other users could maybe take advantage of it too. This PR adds a simple form of support for `locationbias` to the `:google_places_search` lookup.

It can be specified in either the query options or configuration. If both, query takes precedence. A `nil` in the query options overrides a configured value, to omit `locationbias` from the API call, in a similar manner to #1572.

The way I've written it, it only takes a string, e.g. `"point:-36.8509,174.7645"`, which is just passed directly to the API call. I thought about being more sophisticated, _e.g._, getting it to take a hash of the form `{point: [-36.8509, 174.7645]}` instead. But my thinking on this right now is:
- Google Places Search supports four different types of location biases (`ipbias`, `point:lat,lng`, `circle:radius@lat,lng`, `rectangle:south,west|north,east`), each with a different syntax, and it's not obvious a priori how we'd unify them
- If Google adds more types of location biases in the future, it'd be nice for users to be able to use them (even if clunkily) without a gem update. This doesn't mean we should _only_ support strings, but it does mean users should be able to use a string to override whatever smart processing this gem does—so we'd want to support a string anyway, whether or not we also support hashes or other types

That said, very open to suggestions about how to specify this.

Thanks again!


alexreisner:
Sorry for the delay, this looks great! At first, I thought the point should be specified as a two-value array, like it's specified elsewhere in the gem. However, given the different types of bias, I agree that sticking to the API-defined format is probably the best option. Trying to unify this (and maintain the unified syntax as the API evolves) is not worth the trouble. Merging!

PR #1578, opened 2022-05-12
Update Gemfile

jonallured:
Hey @alexreisner I'm looking to contribute to this project but I had a hard time getting setup locally. I was trying to get to a point where I could run `bundle exec rake` and see all the tests pass. My first hurdle was getting mysql installed locally but after some brew commands I was going. I did have to bump to a newer version of that gem - you good with that?

Then my next hurdle was getting a modern version of mongoid installed. For whatever reason this `bson_ext` gem was in my way so I just...removed it? I'm having a hard time seeing the history here, but when I try to find this gem on GitHub I'm not seeing it - maybe it's just really old?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,6 @@ source "https://rubygems.org"
 group :development, :test do
   gem 'rake'
   gem 'mongoid'
-  gem 'bson_ext', platforms: :ruby
   gem 'geoip'
   gem 'rubyzip'
   gem 'rails', '~>5.1.0'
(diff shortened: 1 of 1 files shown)


alexreisner:
Sorry for the inconvenience here. You're not the only one who's had trouble getting set up, so I'm all for making it easier. The `bson_ext` gem is quite old. I'm not remembering the history very clearly but it had something to do with Mongoid. It seems safe to remove it now.

As for Travis, they changed how their support of OS projects works and I haven't been able to get Geocoder back up and running. I've been thinking about using Github's CI features instead but haven't had time to dig into that. (For anyone else who's interested in taking a stab, that would be _hugely_ appreciated.)

PR #1581, opened 2022-05-27
Add github action for running tests

jonallured:
Hey @alexreisner here's my attempt at getting this project back to running on CI! I'm using GitHub Actions here per your comment on that other PR but I'm less familiar with it than say CircleCI so if there's any feedback I'm totally open to changes!

You'll also notice that I had to mark a test as pending because I couldn't figure out why it was failing. You can see that failure here:

https://github.com/jonallured/geocoder/runs/6627303485?check_suite_focus=true

Seems bad to pend but worse to not have CI running at all so I took the tradeoff. What do you think? I also couldn't get other Ruby versions to run but we can iterate from here!

### DIFF
--- a/test/unit/lookups/here_test.rb
+++ b/test/unit/lookups/here_test.rb
@@ -1,5 +1,4 @@
 # encoding: utf-8
-$: << File.join(File.dirname(__FILE__), "..", "..")
 require 'test_helper'
 
 class HereTest < GeocoderTestCase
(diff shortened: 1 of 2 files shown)


test/unit/lookups/here_test.rb
```diff
@@ -1,5 +1,4 @@
 # encoding: utf-8
-$: << File.join(File.dirname(__FILE__), "..", "..")
```
jonallured, line 2: Only this test failed and when I looked at it to see if I could see any differences I noticed this. I removed it wondering if that might fix the issue but it did not. IDK, the mocking/fixture stuff here is a little mysterious to me but I bet it has something to do with load path, right?

alexreisner:
Wonderful! Yes, that test is far less important than having CI restored. Thank you! I'll look into that test.

PR #1584, opened 2022-06-15
Add support for ipbase lookups

jonallured:
Hey @alexreisner here's the PR I've been working on - adding support for ipbase which has seemingly taken over for freegeoip:

```
$ https freegeoip.app
HTTP/1.1 301 Moved Permanently
...snip...
Location: https://ipbase.com/
...snip...
```

I did not delete the freegeoip service in this PR but I suppose that could be done too! Props to my pal @laurabeth for pairing with me on these changes. 🙌 

## Approach

All we really did here was copy/paste code from the freegeoip service and then adapt it to work for ipbase. 🤷 

We started with the lookup class and then added the result class and discovered the way that the tests automatically load in fixture data if you name things right. We worked through the various cases and found the more general tests for lookups and result classes so we made sure we were complying with those requirements too.

## Configuration

I removed some of the configuration that I found - namely supplying a host or protocol because to my understanding these things are not supported. Seemed like simple was better here but open to feedback!

## Demo

We booted up a console and verified that it actually works like so:

```
$ IPBASE_API_KEY=SHHH ./bin/console
> Geocoder.configure(ip_lookup: :ipbase, api_key: ENV["IPBASE_API_KEY"])
=> {:timeout=>3, :lookup=>:nominatim, :ip_lookup=>:ipbase, :language=>:en, :http_headers=>{}, :use_https=>false, :http_proxy=>nil, :https_proxy=>nil, :api_key=>nil, :basic_auth=>{}, :logger=>:kernel, :kernel_logger_level=>2, :always_raise=>[], :units=>:mi, :distances=>:linear, :cache=>nil, :cache_prefix=>nil, :cache_options=>{:prefix=>"geocoder:", :expiration=>nil}, :ipbase=>{:api_key=>"SHHH"}}
irb(main):002:0> results = Geocoder.search("74.33.196.211")
=> [#<Geocoder::Result::Ipbase:0x00007fad92a00950 @data={"data"=>{"timezone"=>{"id"=>"America/Chicago", "current_time"=>"2022-06-15T14:08:06-05:00", "code"=>"CDT", "is_daylight_saving"=>true, "gmt_offset"=>-18000}, "ip"=>"74.33.196.211", "type"=>"v4", "connection"=>{"asn"=>5650, "organization"=>"FRONTIER-FRTR", ...
irb(main):003:0> results.first.country_code
=> "US"
irb(main):004:0> results.first.city
=> "Minneapolis"
> no_results = Geocoder.search("74.33.1909.1908.12312.123")
=> []
```

Did I miss anything? What else should be added when supporting a new ip lookup service?


alexreisner:
...and in light of b05c967569105dcbaa129b143d03dd283f132b30, please call `super` in your test class's `setup` method.

jonallured:
Ok @alexreisner I think I got all the feedback addressed - lmk if you want anything else here!! 🙌

alexreisner:
Looks great, thanks!

PR #1586, opened 2022-07-05
Update yandex lookup error handling

ThinkAndRun:
json response does not contain an error object anymore

https://yandex.ru/dev/maps/geocoder/doc/desc/reference/error_codes.html


alexreisner:
Thanks for this! I also updated the Yandex API response fixture to make the tests pass.

Issue #1568, opened 2022-03-14
geocoder 1.7.4 broken


ollym:
https://github.com/alexreisner/geocoder/commit/c32da2ed4499976c48cf23cb9bb7b951b6852757

Released in geocoder 1.7.4 just broke our app.

We happen to have a top level class called `Google` totally unrelated to geocoder, and also

```ruby
Geocoder.configure(
  lookup: :google
)
```

So tell me - how do I after this patch how do i make sure it picks the `Google` geocoder module instead?

ollym:
The short term fix is to do this in your initialiser:

```ruby
require 'geocoder/lookups/google'

Geocoder.configure(
  lookup: :google
)
```

ollym:
@januszm yes, this is using Rails w/ Zeitwerk

The problem is `Geocoder::Lookup.const_get('Google')` will discover `::Google` if it's defined (which it is in my case) because `require "geocoder/lookups/google"` is no longer being called first.

I suggest you change this

```ruby
def lookup_instance_from(name)
  class_name = classify_name(name)
  begin
    Geocoder::Lookup.const_get(class_name)
  rescue NameError
    require "geocoder/lookups/#{name}"
  end
  Geocoder::Lookup.const_get(class_name).new
end
```

To this:

```ruby
def lookup_instance_from(name)
  class_name = classify_name(name)
  begin
    require "geocoder/lookups/#{name}"
  rescue LoadError
  end
  Geocoder::Lookup.const_get(class_name).new
end
```

januszm:
@ollym  could you please paste the exact error message you're getting?

I could not reproduce this on Rails 7 application with Zeitwerk. I defined an almost empty `Google` class, put `Geocoder.configure(lookup: :google)` and everything went well, I could use both the Geocoder and the Google class.

ollym:
```
NoMethodError
undefined method `new' for Google:Module Geocoder::Lookup.const_get(class_name).new
```

Google in our case is already defined by `google-cloud-trace` gem

ollym:
@januszm @alexreisner Something simple to reproduce it:

```ruby
require 'geocoder'

module Google
end

Geocoder.search('Some Address', lookup: :google)
```

januszm:
So in your case `Google` is a module. I still can't reproduce this, with the debugger inside the gem code:

```ruby
> Geocoder.search('Some Address', lookup: :google)

[...]

PR #1588, opened 2022-08-01
existing constant name shadowing a lookup class name results in NoMethodError

avram-twitch:
Related issue here: https://github.com/alexreisner/geocoder/issues/1568

It was marked as close and resolved, but I'm still seeing the same problem. I was able to recreate the bug on ruby `2.6.3` and geocoder version `1.8.0` with the following

```
require 'geocoder'

module Google
end

Geocoder.configure(:api_key => "YOUR-KEY", :lookup => :google)
Geocoder.search("Something")
```

Results in

```
        6: from test.rb:7:in `<main>'
        5: from /Users/avramtwitchell/.rvm/gems/ruby-2.6.3/gems/geocoder-1.8.0/lib/geocoder.rb:22:in `search'
        4: from /Users/avramtwitchell/.rvm/gems/ruby-2.6.3/gems/geocoder-1.8.0/lib/geocoder/query.rb:11:in `execute'
        3: from /Users/avramtwitchell/.rvm/gems/ruby-2.6.3/gems/geocoder-1.8.0/lib/geocoder/query.rb:40:in `lookup'
        2: from /Users/avramtwitchell/.rvm/gems/ruby-2.6.3/gems/geocoder-1.8.0/lib/geocoder/lookup.rb:108:in `get'
        1: from /Users/avramtwitchell/.rvm/gems/ruby-2.6.3/gems/geocoder-1.8.0/lib/geocoder/lookup.rb:121:in `spawn'
/Users/avramtwitchell/.rvm/gems/ruby-2.6.3/gems/geocoder-1.8.0/lib/geocoder/lookup.rb:146:in `instantiate_lookup': undefined method `new' for Google:Module (NoMethodError)
```

I added the `inherit=false` flag to the `const_get` call, which prevents looking up constants higher up in the module namespace. I also added a test that fails with the code as it was. I'm unfamiliar with using the `test-unit` gem for testing, so feel free to make suggestions!

### DIFF
--- a/lib/geocoder/lookup.rb
+++ b/lib/geocoder/lookup.rb
@@ -138,9 +138,9 @@ module Geocoder
     # Safely instantiate Lookup
     #
     def instantiate_lookup(name)
-      class_name = "Geocoder::Lookup::#{classify_name(name)}"
+      class_name = classify_name(name)
       begin
-        Geocoder::Lookup.const_get(class_name)
+        Geocoder::Lookup.const_get(class_name, inherit=false)
       rescue NameError
         require "geocoder/lookups/#{name}"
       end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -772,3 +772,6 @@ class MockHttpResponse
     @headers[key]
   end
 end
+
+module Google
+end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -773,5 +773,5 @@ class MockHttpResponse
   end
 end
 
-module Google
+module MockLookup
 end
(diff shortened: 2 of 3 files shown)


avanrielly:
@alexreisner Is there any chance of this getting merged?  We just updated to 1.8.0 and we are getting this same error.

alexreisner:
Sorry, this somehow slipped through the cracks. Reviewing now, will merge shortly. Will also release 1.8.1 soon.

avanrielly:
Thanks for the quick response.

PR #1591, opened 2022-09-16
Allow geoapify lookup to perform autocomplete query

Intrepidd:
Geoapify has an [autocomplete api](https://www.geoapify.com/address-autocomplete)

This PR allows to use either the search or autocomplete API


alexreisner:
Looks great, thank you!

Intrepidd:
Fabulous ! Do you think we could get a release ?

PR #1599, opened 2023-01-05
Rails geocoder - Deprecation Warning

FlandersBurger:
Resolves the deprecation warning when supplying a hash to Geocoder.search.
Search checks whether the supplied search is_ipaddress? which in turn goes to the below function and displays the warning

`/.rbenv/versions/2.7.2/lib/ruby/gems/2.7.0/gems/geocoder-1.8.1/lib/geocoder/ip_address.rb:13: warning: Using the last argument as keyword parameters is deprecated`

i.e.
`Geocoder.search({ :city=>"Victoria" })`

### DIFF
--- a/lib/geocoder/ip_address.rb
+++ b/lib/geocoder/ip_address.rb
@@ -21,7 +21,7 @@ module Geocoder
     end
 
     def loopback?
-      valid? and !!(self == "0.0.0.0" or match(/\A127\./) or self == "::1")
+      valid? and !!(self == "0.0.0.0" or self.match(/\A127\./) or self == "::1")
     end
 
     def private?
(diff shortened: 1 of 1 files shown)


FlandersBurger:
I haven't submitted a PR to a public codebase before so I would much appreciate knowing what I need to change :)

Cheers

alexreisner:
Thank you! This is great.

As for PR feedback, it's all positive. As someone who reviews a lot of these, I especially appreciate:

* your succinct explanation of the problem, with enough detail to reproduce quickly
* your commits (after being squashed) only change what needs to be changed to address the issue
* you did not include any project-specific (irrelevant to Geocoder) details

I will merge this shortly.

PR #1601, opened 2023-02-06
 Add connect_timeout to configuration

ybregey:
This PR allows `open_timeout` to be configurable. I have a case where it is critical to quickly switch to another geocoding provider if the primary one is unavailable or responds too slowly.

Failed requests caused by exceeding `open_timeout` or `timeout` will be handled the same way thanks to [the current error handling approach](https://github.com/alexreisner/geocoder/blob/059bf127fe5967d013783faf01586504b2fe1b6f/lib/geocoder/lookups/base.rb#L311-L312).
image


ybregey:
@alexreisner sorry for pinging again. any chance this could be merged?

ybregey:
Thanks for the suggestions 🙌  Updated.

Issue #934, opened 2015-10-26
Add support to MSSQL/TSQL (Azure)


ipepe:
Hello.
Im running rails app on Jruby with MSSQL database but geocoder deosnt support that. Queries have ATAN2 instead of ATN2 of MSSQL. If You need a database I can setup one in azure for the matter of working on that. I You need my help, just tell me.

alexreisner:
Sorry, but I feel similarly about MSSQL as about Oracle (see #535). These databases are not especially popular in the Ruby community and the effort required to add compatibility is quite significant. Only way I could consider adding is if someone submitted a clean pull request (minimal code changes) that included the necessary updates to Travis CI config. Even then, I'd really have to weight the depth and complexity of the code changes against the benefits.

ipepe:
What about doing it like sqlite?

alexreisner:
I don't like how SQLite is handled currently. Ideally there would be separate drivers for each DB system. The way SQLite is currently implemented it's a special case, handled mostly by `if` statements. I don't like the idea of introducing more branches (because it's not good code) or restructuring the whole thing (better code, but deep changes) for functionality that's not core.

ipepe:
based on this: 
http://stackoverflow.com/questions/13026675/calculating-distance-between-two-points-latitude-longitude
I declare that I want to add support to MSSQL. I will create pull request. But before I dive into code, is there a chance for those separate adapters, or maybe just a stub of that?

alexreisner:
OK, you're welcome to submit a PR but just to be clear: there is a good chance I won't merge because I can't currently think of an implementation that I feel comfortable with. Even if not merged, your work will still be useful because you'll be able to install the gem from your own repo. Just wanted to be clear about this.

Issue #532, opened 2013-10-08
SqlServer - ATAN2 bearing calculation fails


mmmries:
I couldn't find any existing issues filed for this. Sorry if it is a duplicate.

I'm using SqlServer at work (grudgingly) and using the standard SomeModel.near method results in this error:

``` ruby
ActiveRecord::StatementInvalid: TinyTds::Error: 'ATAN2' is not a recognized built-in function name.
```

It looks like it is used to calculate the bearing. I am able to work around the problem by using something like:

``` ruby
MyModel.near('eiffel tower', 20.0, select_bearing: false)
```

But I had to dig through the source a bit for this. Maybe we should add some documentation about sql server or else change the default select_bearing option to be false when using TinyTds?

slead:
I'm also using SQL Server, and am seeing an error `undefined method `to_f'` even when I use `select_bearing: false`:

```
@parcel = Parcel.first
@parcel.geocoded?
  => true
Cpt.first.geocoded?
  => true
Cpt.near(@parcel)
  => ActiveRecord::StatementInvalid: TinyTds::Error: 'ATAN2' is not a recognized built-in function name
Cpt.near(@parcel, select_bearing: false)
  => NoMethodError: undefined method `to_f' for {:select_bearing=>false}:Hash
```

Thanks for any advice.

ttilberg:
I came across this issue today and worked around it as described below. This is old, but I wanted to leave a note here since this is directly relevant to this issue.

I'm not sure if Alex would be open to a pull request to add such functionality, but generally speaking, the are two changes that should happen to the bearing related sql make this more smooth in SQL Server:

- references to `ATAN2` should be `ATN2` when SQL Server
- references to `MOD(a,b)` should be rewritten as `A % B`

For my quick fix, I just overwrote the method. But if it were desirable to improve SQL Server support for this gem, here are my notes:

There are already methods for detecting postgres and sqlite, so adding sqlserver detection wouldn't be too intrusive (reference `/tinytds|mssql|sqlserver/`). Being that the bearing SQL is just a string, it would be easy to template the ATAN2 function from a variable, like `sql_atan2` (not sure where this would be best implemented). Regarding modulo, I see that postgres also supports the `a % b` syntax, but I'm not sure what else this gem supports or it's impact. It could be reasonable to simply rewrite the SQL using the `%` operator.

Here is how I rewrote the method to accommodate SQL server. I added this to my initializers and it seems to work.

[...]

PR #1605, opened 2023-02-24
Refactor: Support for SQL Server (WIP for discussion)

doconnor-clintel:
Previously:

- #934
- #532
- https://gist.github.com/ttilberg/fe5ac60351ad33b865550c08857f612f via @ttilberg 

## Problem

Right now, SQL Server has bespoke SQL required to do bearing calculations.
Other datastores have a similar problem, ie SQLite, depending on what is available.

At some point, Geocoder::Sql was split out to try to make managing this a little bit easier, but it accidentally lead to the situation where:

* `Geocoder::Store::Activerecord` had a number of methods and conditionals to decide which of the other class' methods to call
* `Geocoder::Sql` doesn't know what database it is connected to, so can't generate correct SQL for every situation

## Recommended solution

Ultimately, I think the responsibility for which datastore am I connected to does live in `Geocoder::Store::Activerecord`; *but* specific backends should be available for standard sql, vs sqlite, vs postgres, vs sql server.

This might seem like it violates DRY, but ultimately it would help acheive a single-responsibility-principle.

### Transitioning and backwards compatibility

What I have done in this PR is really a step 1; where the standalone methods are marked deprecated.
This allows a future major version to remove them.

Additionally, at least for now I have shifted the sql generating concerns back into Geocoder::Store::Activerecord. This is because there is an implicit coupling between "which SQL should I generate" and "what am I connected to?". It introduces an extra layer of conditionals in the sql generation temporarily.

A subsequent PR should (with a major version to indicate the BC break)

* Implement Geocoder::Store::ActiveRecord::Postgres
* Implement Geocoder::Store::ActiveRecord::SQLServer
* Implement Geocoder::Store::ActiveRecord::SQLite
* Implement Geocoder::Store::ActiveRecord::SQLiteWithExtensions
* Implement Geocoder::Store::ActiveRecord::Postgis?
* Implement Geocoder::Store::ActiveRecord::OGC or SQL/MM if relevant?

### DIFF
--- a/lib/geocoder/sql.rb
+++ b/lib/geocoder/sql.rb
@@ -11,6 +11,8 @@ module Geocoder
     # http://www.scribd.com/doc/2569355/Geo-Distance-Search-with-MySQL
     #
     def full_distance(latitude, longitude, lat_attr, lon_attr, options = {})
+      warn "Direct use of this method is deprecated, prefer Geocoder::Store::ActiveRecord.full_distance_sql"
+
       units = options[:units] || Geocoder.config.units
       earth = Geocoder::Calculations.earth_radius(units)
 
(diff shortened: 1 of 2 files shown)


alexreisner:
@doconnor-clintel Thanks for this and sorry for the delay. Of course you're right. The current "design" comes from many years of doing things quickly and simply. So I'm open to refactoring. This is a big project. If you're willing to take the lead, let's get started.

One thing to consider is that it's actually not necessary to deprecate anything. Methods in `Geocoder::Sql` could call methods in `Geocoder::Store::ActiveRecord`. Could we implement it that way first, and then deprecate the methods later? That will make it easier to release.

PS: `ATN2`. Wow.

PR #1606, opened 2023-02-24
Disable inherit flag const_get in `Geocoder::Cache#initialize`

shqear93:
Based on [this](https://github.com/alexreisner/geocoder/commit/4b9d2c689e8cacb6ce71714a7271bdb0952761cd) conversation, I think we should disable inheritance lookup behaviour for `get_const` method


alexreisner:
I agree. Thank you!

PR #1607, opened 2023-02-26
Update `README.md` file, use DB as storage

shqear93:
@alexreisner I would like to hear your input about this, this is very optional addtition to the reademe file that might be helpful for some users

### DIFF
--- a/README.md
+++ b/README.md
@@ -511,6 +511,13 @@ If you store multiple sets of coordinates for each object, you can specify latit
 Venue.near("Paris", 50, latitude: :secondary_latitude, longitude: :secondary_longitude)
 ```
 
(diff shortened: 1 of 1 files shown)


alexreisner:
@shqear93 I like this. Two suggestions:

- This should go in the "Caching" section, right above the last line, which is in italics ("Before you implement caching...").
- I'm trying to reduce the amount of text in the README. Can this be a single line?: "In addition to conventional cache stores like Redis, it's possible to keep your cache in the database using ActiveRecord. For example see [this gist](https://gist.github.com/shqear93/4b07153b4ca7e4e4a41da492679f6c0e)."

shqear93:
@alexreisner done 🆗

PR #1609, opened 2023-03-03
Feature: Include region on AWS lookup client config

3zcurdia:
# Changelog

- Include region on Aws::LocationService::Client config
- Remove region key when is nil 

## Why

To avoid setup region via ENV vars while its using the keys approach


alexreisner:
Thanks!

Issue #1610, opened 2023-03-14
ESRI lookup no error handling


acdlbs:
### Expected behavior
When using geocoder with ESRI, 
if I make a request and there is an error returned,
I would expect an exception to be raised and to see the error.

![image](https://user-images.githubusercontent.com/3392886/225144703-431704f4-b9d8-417f-8b00-a52658f50931.png)

### Actual behavior
When using geocoder with ESRI, 
if I make a request and there is an error,
I am returned an empty array.

![image](https://user-images.githubusercontent.com/3392886/225147176-50b60e5e-80ea-4f71-8099-adffd2421a87.png)

* Geocoder version: Geocoder 1.8.0

PR #1611, opened 2023-03-14
Updated esri lookups error handling

acdlbs:
Updated error handing for the `results` method in `lookups/esri.rb`

More inline with other geocoding `results` methods - see here: https://github.com/alexreisner/geocoder/blob/1928f7886823d63e0268eebd693ee20314ebdb08/lib/geocoder/lookups/bing.rb#L36

Before when ESRI would fail on a geocode (an example would be an invalid api key) Geocoder would return an empty array instead of an exception. 

Attached issue - https://github.com/alexreisner/geocoder/issues/1610


alexreisner:
Excellent, thank you!

PR #1616, opened 2023-05-25
Update README to document services that do not support caching

amleaver:
I was recently working on a Geocoder integration using the Amazon Location Service backend but was confused by why lookups weren't being cached. 

I've updated the README to reflect the services which do not appear to support caching to help future developers. [I've based this list on an existing test.](https://github.com/alexreisner/geocoder/blob/de6be6790646accbe6e377145ea60a7a3c1d5730/test/unit/cache_test.rb#L19)

I'm unclear if caching is not permitted with AWS, but if there is an desire to add it then I can look into raising a PR with support for it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -378,6 +378,12 @@ Do *not* include the prefix when passing a URL to be expired. Expiring `:all` wi
 
 _Before you implement caching in your app please be sure that doing so does not violate the Terms of Service for your geocoding service._
 
+The following services do not support caching:
+
+* Amazon Location Service (`:amazon_location_service`)
+* GeoLite2 (`:geoip2`)
+* MaxMind Local (`:maxmind_local`) 
+
 
 Advanced Model Configuration
 ----------------------------


alexreisner:
Thanks for this. I did not realize caching wasn't supported with the Amazon lookup but I now see that it's bypassing the `fetch_data` method so, yeah, no caching. I'm open to a PR that would add cache support, preferably by making it more like other lookups (see `:google` and `:nominatim` for good examples) and calling `fetch_data`, but I know Amazon works a bit differently, so there may be limits on how similarly it can be structured.

As for this README change, I'm worried it won't be maintained, since it's separate from [the rest of the API-specific info](https://github.com/alexreisner/geocoder/blob/master/README_API_GUIDE.md). I think if we want to document lack of caching it should probably be mentioned in the "Limitations" section of the lookup in question. Also, once Amazon supports caching, there may be no need for this, since the other two non-caching lookups are local.

amleaver:
Thanks for the quick response, and that's a reasonable concern, I was on the fence about which of the two to update in the first place. 

I've updated the API guide limitations and made a note to check this in the main README.

I experimented briefly with adding caching to the existing service and it can be done without necessarily needing to use `fetch_data` as the response from the AWS SDK is a series of Structs which can be easily serialized and stored in the cache.

That said it's arguably a more brittle and error prone approach than simply caching JSON.

alexreisner:
Thanks, this looks good. I'll merge shortly.

As for adding caching for AWS, I'd strongly prefer to cache the JSON and stick with the standard methods that are used by the other lookups (eg, `fetch_data`), so features (like caching) can be implemented globally. I'm open to arguments why this one warrants an exception, though, if it seems like it really does.

PR #1618, opened 2023-06-12
request metadata in Trimble maps single search call

arianseyeditfx:
To add useful metadata to the response data for Trimble maps single search integration we need to add the
[`include` url parameter with value 'Meta'](https://developer.trimblemaps.com/restful-apis/location/single-search/single-search-api/).

This PR takes care of such addition.

example call
``` https://singlesearch.alk.com/na/api/search?authToken=<TOKEN>&include=Meta&query=Feasterville Trevose, Pennsylvania 19053, United States```

response:
```{
    "Err": 0,
    "ErrString": "OK",
    "QueryConfidence": 1,
    "TimeInMilliseconds": 125,
    "GridDataVersion": "GRD_ALK.NA.2023.01.18.29.1.1",
    "CommitID": "pcmws-22.08.11.0-1778-g586da49bd1b: 05/30/2023 20:14",
    "Locations": [
        {
            "Address": {
                "StreetAddress": "",
                "LocalArea": "",
                "City": "Feasterville Trevose",
                "State": "PA",
                "StateName": "Pennsylvania",
                "Zip": "19053",
                "County": "Bucks",
                "Country": "US",
                "CountryFullName": "United States",
                "SPLC": null
            },
            "Coords": {
                "Lat": "40.150025",
                "Lon": "-75.002511"
            },
            "StreetCoords": {
                "Lat": "40.150098",
                "Lon": "-75.002827"
            },
            "Region": 4,
            "POITypeID": 0,
            "PersistentPOIID": -1,
            "SiteID": -1,
            "ResultType": 4,
            "ShortString": "Feasterville Trevose, PA, US, Bucks 19053",
            "GridID": 37172748,
            "LinkID": 188,
            "Percent": 6291,
            "TimeZone": "GMT-4:00 EDT"
        }
    ]
}
```
Important Note:
- Top level metadata "QueryConfidence" from the above would **need** to be inserted in the results array that would include the Result object (in this case Geocoder::Result::PcMiler) with service specific data, so that it could be available for the user. This PR also takes care of that.

- all Fixtures are also updated for consistency sake

### DIFF
--- a/lib/geocoder/lookups/pc_miler.rb
+++ b/lib/geocoder/lookups/pc_miler.rb
@@ -35,9 +35,17 @@ module Geocoder::Lookup
     def results(query)
       return [] unless data = fetch_data(query)
       if data['Locations']
+        add_metadata_to_locations(data)
         data['Locations']
       else
         []
+      end
+    end
+
+    def add_metadata_to_locations(data)
+      confidence = data['QueryConfidence']
+      data['Locations'].each do |location|
+        location['QueryConfidence'] = confidence
       end
     end
 
--- a/lib/geocoder/lookups/pc_miler.rb
+++ b/lib/geocoder/lookups/pc_miler.rb
@@ -35,14 +35,14 @@ module Geocoder::Lookup
     def results(query)
       return [] unless data = fetch_data(query)
       if data['Locations']
-        add_metadata_to_locations(data)
+        add_metadata_to_locations!(data)
         data['Locations']
       else
         []
       end
     end
 
-    def add_metadata_to_locations(data)
+    def add_metadata_to_locations!(data)
       confidence = data['QueryConfidence']
       data['Locations'].each do |location|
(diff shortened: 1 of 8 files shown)


arianseyeditfx:
@alexreisner could you take a look at this PR for a review? Thank you so much

alexreisner:
Hmm, @alexdean we have not seen this with other adapters, unless I'm forgetting one. But I agree that there's no way around it. This looks good to me with just one nitpicking suggestion: can we change the `add_metadata_to_locations(data)` method name to `add_metadata_to_all_locations!(full_response)`. The bang is a convention used throughout the gem for methods with side effects, `full_response` helps me understand what the data is, and `all_locations` explains more clearly what's happening. Does that all make sense?

Thanks for this, by the way.

arianseyeditfx:
@alexreisner add_metadata_to_locations method was changed to add_metadata_to_locations!

Thanks for the suggestion

PR #1619, opened 2023-06-13
Fix : param type in ban_data_gouv_fr lookup

guillaumj:
Updates for the ban_data_gouv_fr lookup : 
- Remove the old `village`, `town` and `city` (result) types which are no longer used
- Add the new `municipality` type (replacing old types)
- Update API Guide urls

Doc : https://adresse.data.gouv.fr/api-doc/adresse

Capture d’écran 2023-06-14 à 01 08 34


alexreisner:
This lookup hasn't been touched in a while. Thanks for this! Looks good. Will merge.

PR #1622, opened 2023-06-29
Remove double encoding

karina:
Remove double encoding because `hash_to_query` already does this in the base class.

### DIFF
--- a/lib/geocoder/lookups/pc_miler.rb
+++ b/lib/geocoder/lookups/pc_miler.rb
@@ -52,14 +52,14 @@ module Geocoder::Lookup
     def query_url_params(query)
       if query.reverse_geocode?
         lat,lon = query.coordinates
-        escaped_query = "#{CGI.escape(lat)},#{CGI.escape(lon)}"
+        formatted_query = "#{CGI.escape(lat)},#{CGI.escape(lon)}"
       else
-        escaped_query = CGI.escape(query.text.to_s)
+        formatted_query = query.text.to_s
       end
 
       {
         authToken: configuration.api_key,
-        query: escaped_query,
+        query: formatted_query,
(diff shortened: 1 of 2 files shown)


alexreisner:
Thanks for this! It looks good except it's causing two other tests to fail, where `%20` is expected in the URL instead of `+`. Either one should work, so I think it's fine to change the expected URL in the other tests. (@alexdean does this make sense to you?)

karina:
@alexreisner Just updated the test! We noticed this was causing issues with querying PC Miler with queries that had `,`.

alexdean:
@alexreisner yes makes sense to me.

fwiw i was following what the Mapbox lookup code does.  ([code reference](https://github.com/alexreisner/geocoder/blob/a692545419f2e8fcc613c70bf6fb05edb8dc0aa3/lib/geocoder/lookups/mapbox.rb#L40))

is that a bug also? i may have just misunderstood some part of the process, too...

alexreisner:
@alexdean good question. There's a similar test for the Mapbox lookup that is passing:

https://github.com/alexreisner/geocoder/blob/a692545419f2e8fcc613c70bf6fb05edb8dc0aa3/test/unit/lookups/mapbox_test.rb#L15

I think the difference is that, in the Mapbox lookup, the output of that method is put directly into the URL, so it bypasses the normal Geocoder flow where the query string is escaped:

https://github.com/alexreisner/geocoder/blob/a692545419f2e8fcc613c70bf6fb05edb8dc0aa3/lib/geocoder/lookups/mapbox.rb#L14

So basically you got unlucky in the lookup you chose to copy/paste from--ha! In general, I wish this process of adding new lookups was a bit easier. It's a weakness of the gem.

@karina thanks for the update! I will merge (and deploy new version in the next day or two).

PR #1626, opened 2023-07-03
feat: Provide better Rails cache store support

ixti:
Current implementation disrespects expiration TTL, polluting redis with non-expirable keys. Proposed solution wraps Rails.cache (configured with :redis_cache_store) with adapter that respects Rails cache semantics.


ixti:
It's also worth to mention in the README that configuration like:

``` ruby
Geocoder.configure(
  cache: ...,
  cache_options: { expiration: 1.week },
  ...
)
```

Sets prefix to `nil`, not the default `geocoder:`. We had configuration like:

``` ruby
cache: Geocoder::CacheStore::Generic.new(Rails.cache, {}),
cache_options: { expiration: 1.week },
```

As a result we have lots of keys both without prefix and without expiration.

PR #1629, opened 2023-08-16
Remove references to legacy Google Places Search fields

### LINKED ISSUE #1627: Unsupported field name 'id' error when using Google Places Search API
When attempting to perform a Places Search API lookup for a given location using the default fields specified by the gem, no results are returned, and an error is printed to the console due to the use of deprecated and now removed `id` and `reference` fields.

### Expected behavior

The default fieldset specified by the gem should be valid.

### Actual behavior

The following error is printed and no results are returned:

```
Google Places Search API error: invalid request (Error while parsing 'fields' parameter: Unsupported field name 'id'. ).
```

### Steps to reproduce
1. Make a `search` request with `lookup: :google_places_search` without specifying `field:`

### Environment info

* Geocoder version: 1.8.2
* Rails version: 7.0.5.1
* Database (if applicable): N/A
* Lookup (if applicable): `google_places_search`

BytewaveMLP:
Removes legacy `id` and `reference` fields from Google Places Search results.

Closes #1627.

### DIFF
--- a/lib/geocoder/lookups/google_places_search.rb
+++ b/lib/geocoder/lookups/google_places_search.rb
@@ -49,12 +49,11 @@ module Geocoder
       end
 
       def default_fields
-        legacy = %w[id reference]
         basic = %w[business_status formatted_address geometry icon name 
           photos place_id plus_code types]
         contact = %w[opening_hours]
         atmosphere = %W[price_level rating user_ratings_total]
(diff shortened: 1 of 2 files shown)


alexreisner:
Thanks! This is perfect. Can you remove those two fields from the test (see failure)? Then we should be good to merge.

alexreisner:
Excellent, thanks!

PR #1630, opened 2023-08-24
Add 1.8.2 to CHANGELOG.md in master

jaredbeck:
Copied from tag v1.8.2. Somehow this changelog entry never made it to master?


alexreisner:
Haha! Wow, good catch. It looks like I forgot to push from my local machine after releasing. I just pushed [b546dd6](https://github.com/alexreisner/geocoder/commit/b546dd637672b9bf84abf692b4594f39cdc0ad48) on master, which also bumps the version number. Thank you for this.

PR #1633, opened 2023-09-14
Add IP2Location LITE support

ip2location:


ip2location:
@alexreisner May we know how to fix the timeout error?

ip2location:
Hi @alexreisner, can you guide us on how to solve the timeout error?

alexreisner:
Yes, sorry for the delay. I think the thing to do here is to add `:ip2location_lite` to the array on [this line](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookup.rb#L26), to specify that the lookup doesn't make HTTP requests. (It looks like `:maxmind_local` should be there too, but I'll handle that separately.)

ip2location:
@alexreisner We have updated https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/lookup.rb#L26

alexreisner:
Thanks! I'm seeing another error, which I'm looking into. More on that shortly.

For now, two thoughts:

* Please be sure to call `super` at the beginning of the `setup` method in the test file.
* When I add the `ip2location_ruby` gem, I get `NoMethodError: private method 'open' called for #<Geocoder::Lookup::Ip2location:0x0... @cache=nil>`.

alexreisner:
I think the best way to handle the error is to add `gem 'ip2location_ruby'` at line 10 of the Gemfile.

ip2location:
@alexreisner We have fixed the error.

ip2location:
Hi @alexreisner, can you help to review this PR?

alexreisner:
Looks great. Thanks!

PR #1634, opened 2023-09-19
adding support for PDOK NL

opensourceame:
This PR adds support for [pdok](https://www.pdok.nl), a new Geocoding service for The Netherlands that replaces Nationaal Georegister.


.gitignore
```diff
@@ -6,3 +6,4 @@ rdoc/*
 Gemfile.lock
 api_keys.yml
 test/geocoder_test.sqlite
+.idea/*
```
opensourceame, line 9: `.idea` is created by RubyMine which is a very popular IDE

opensourceame:
@alexreisner the old Nationaal Georegister has been deprecated. I can remove the code from the old platform in this PR or a create a new PR for that. What would you prefer?

.gitignore
alexreisner: Yes, however, everything in this `.gitignore` file should be specific to the Geocoder gem. To ignore files related to local development, please use your `~/.gitignore_global` file.

opensourceame:
> Thanks for this! It looks great. If you can remove that line from `.gitignore`, I'll merge. And feel free to submit another request for removing the deprecated API.

Done @alexreisner 👍

PR #1637, opened 2023-11-06
Add lookup for Google Places Details API (New)

patchy-oss:
## Overview
Adds a lookup for Google Places Details (New).

## Changes
- Added new base class for Google - `GoogleBase`.
- Added a base class for new Google API - `GoogleNew`.
- Added new lookup for Google places details (new).
- Tests.

## Notes
Most of the changes are copy-paste with small corrections from the existing google lookups. I've tried to unify both google APIs, but seems like there's not too much to unify... Also I think the `GoogleBase` class is kinda redundant, so I can duplicate `configure_ssl!` and `supported_protocols` methods in both `Google` and `GoogleNew` instead (or maybe you have a better solution).
Naming is tough 🙃 

Google documentation:
- [Place Details (New)](https://developers.google.com/maps/documentation/places/web-service/place-details)
- [REST Resource: places](https://developers.google.com/maps/documentation/places/web-service/reference/rest/v1/places)
- [Migrate the Places API response](https://developers.google.com/maps/documentation/places/web-service/migrate-response)

### DIFF
--- a/lib/geocoder/lookup.rb
+++ b/lib/geocoder/lookup.rb
@@ -38,6 +38,7 @@ module Geocoder
         :google_premier,
         :google_places_details,
         :google_places_search,
+        :google_places_details_new,
         :bing,
         :geocoder_ca,
         :yandex,
--- a/lib/geocoder/lookups/google.rb
+++ b/lib/geocoder/lookups/google.rb
@@ -1,8 +1,8 @@
-require 'geocoder/lookups/base'
+require 'geocoder/lookups/google_base'
 require "geocoder/results/google"
 
 module Geocoder::Lookup
-  class Google < Base
+  class Google < GoogleBase
 
     def name
       "Google"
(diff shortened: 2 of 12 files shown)


maokomioko:
@alexreisner Hi Alex, possible to merge you think?

zjwhitehead:
@alexreisner do you mind taking a look at this? 
It would be handy to have this support as Google pushes users to upgrade to this API.

alexreisner:
Really sorry for the massive delay here. I fully intend to review and merge this. However, I think a big of restructuring is needed. (@patchy-oss if you're still interested, feel free to do this, or someone else can take it on.)

Basically, if the new Google API supersedes the old one (ie, the old one is deprecated), we should replace the current GooglePlacesDetails classes, rather than adding GooglePlacesDetailsNew. If it can't be backward-compatible with the old API, that's fine, and it will be released as part of a new minor version of Geocoder.

Ideally, we should avoid adding a GoogleBase class unless it's really usable by **all** GoogleX lookups. But even then, I'd prefer if that were added in a separate PR. Too many changes make things hard to review, which is a tiny bit of the reason for the delay on this.

Does that make sense? Thanks for your patience.

PR #1639, opened 2023-12-15
[docs] add known issues about near scope

takatea:
Thank you for your fantastic gem!

I met errors or a performance degradation when I tried to use Rails's `pluck` or `select` with the near scope of geocoder.
Finally, I found the solution with select option of near scope,  but unfortunately it is not documented well in the README.
So, this PR adds descriptions for such cases.

By the way, I have some ideas to make select option more friendly for Rails users.
I will open issues for them.


README.md
```diff
@@ -771,6 +775,119 @@ Known Issues
...
+The above SQL has only necessary columns.
+
+#### Using `near` with `pluck`
+
+You cannot use [ActiveRecord's pluck method](https://api.rubyonrails.org/classes/ActiveRecord/Calculations.html#method-i-pluck) after `near` scope.
+
+For example, the following code:
+
+```ruby
+
```
kou, line 861:
```suggestion
```

alexreisner:
Thanks for this. It's very helpful. However, it's a lot of additional information, and I'm trying to keep the README as short as possible (it's already too long). Do you think we can get rid of the new sections and limit any additions to 10 lines? The problem is explained well enough on [line 735](https://github.com/alexreisner/geocoder/blob/master/README.md?plain=1#L735), so I don't think we need the SQL examples. The most important things are the workarounds: the `:select` option on the `near` method and `to_a.pluck`. What if we just added the workarounds?

README.md
takatea: @kou 
Thanks for your review!
I will fix your suggestion.

takatea:
@alexreisner 

Thanks for your review!
I did not realise the section you mentioned.
I agree to keep the README simple, and I will update this PR based on your feedback!

README.md
takatea: Fixed in commit: https://github.com/alexreisner/geocoder/pull/1639/commits/b1debc0eb50e3224a834a68ad4d14041537acb97

takatea:
Hi, @alexreisner !

I fixed the document based on your suggestion, but it exceeds the 10-line limit slightly. 
It seems we can remove the known issue related to `includes` ([line774-792](https://github.com/takatea/geocoder/blob/cdb47eeb58187842cbc3c18aca1cd6e0380e1dc4/README.md?plain=1#L774-L792)) . What do you think?

Please review, and let me know if any adjustments are needed!
Thanks in advance.

takatea:
@alexreisner 

Hi there!
I pushed changes based on your feedback.
Could you review this updated PR?

alexreisner:
Hi! Sorry, I lost track of this. It looks great. Thank you!

Issue #1642, opened 2024-01-01
base64 and csv will be bundled gem in Ruby 3.4


JunichiIto:
### Expected behavior

No warnings.

### Actual behavior

I got the following warnings about base64 and csv when running geocoder tests in Ruby 3.3.0:

```
/Users/jnito/dev/oss/geocoder/lib/geocoder/lookups/google_premier.rb:2: warning: base64 was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add base64 to your Gemfile or gemspec.
/Users/jnito/dev/oss/geocoder/lib/geocoder/lookups/maxmind.rb:3: warning: csv was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add csv to your Gemfile or gemspec.
```

Please read the news below:
https://github.com/ruby/ruby/blob/v3_3_0/NEWS.md#stdlib-updates

> RubyGems and Bundler warn if users do require the following gems without adding them to Gemfile or gemspec. This is because they will become the bundled gems in the future version of Ruby.

### Steps to reproduce

```
bundle exec rake
```

### Environment info

* Geocoder version: master
* Rails version: N/A
* Database (if applicable): 
* Lookup (if applicable):

PR #1643, opened 2024-01-01
Fix dependency warning in Ruby 3.3 and prepare for Ruby 3.4

JunichiIto:
Fix https://github.com/alexreisner/geocoder/issues/1642


alexreisner:
Thanks for this! I will merge, but first, one thing. Because these gems are only required for specific lookups, we might want to indicate that with a comment at the end of each line (like `# for :google_premier lookup`). That way, if the lookups are removed in the future, we know we can remove the dependency too. Does that make sense?

JunichiIto:
Thank you for your review. I added comments.

PR #1644, opened 2024-01-08
Add place_id as a result attribute from amazon_location_service

astoica1986:
Place_id is a really useful unique location identifier. Since at least 2 of Amazon Location Service's place providers return this attribute, we would like to return it as a an attribute in the geocoder response as well.


alexreisner:
Sorry for the delay here, and thanks!

PR #1646, opened 2024-01-11
Add support for GetAddress (getaddress.io) postcode geocoding and lookups

mark-young-atg:


alexreisner:
Thanks for this, and sorry for the delay. I'm having trouble trying this out. Every `Geocoder.search` request seems to return an empty array. Possible reasons:

- Request URL includes `v2/uk` when it looks from the docs like it should be `location` instead.
- Words in the search query seem to get concatenated, without spaces, which might not be what you intended.

mark-young-atg:
@alexreisner Thank you for your comments. You are right, the documentation has moved on. We are using the API as it previously worked, so for instance:

```
https://api.getaddress.io/v2/uk/M14AL?api-key=<key>
{
    "Latitude": 53.479854583740234,
    "Longitude": -2.238126277923584,
    "Addresses": [
        "Post Office, Unit F4, Piccadilly Plaza, , , Manchester, ",
        "Unit F4, Piccadilly Plaza, , , , Manchester, "
    ]
}
```

We remove the spaces for the postcodes, I presume that was required at some point, although it does not appear to matter now.

It seems that getAddress now want you to go through a two step process to get to the Latitude and Longitude, e.g.

```
https://api.getaddress.io/location/M1 4AL?api-key=<key>
{
    "suggestions": [
        {
            "location": "M1 4AL, Manchester",
            "url": "get-location/p_ODg0MjgyZGRkYjQzMzNkIDgyNDA0MyA0YzRmMmFlMzIyNWQ0NmM=",
            "id": "p_ODg0MjgyZGRkYjQzMzNkIDgyNDA0MyA0YzRmMmFlMzIyNWQ0NmM="
        }
    ]
}

and then use the provided `url`/`id` value in the next query

https://api.getaddress.io/get-location/p_ODg0MjgyZGRkYjQzMzNkIDgyNDA0MyA0YzRmMmFlMzIyNWQ0NmM=?api-key=<key>
{
    "latitude": 53.479854583740234,
    "longitude": -2.238126277923584,
    "area": "",
    "town_or_city": "Manchester",
    "country": "England",
    "county": "",
    "postcode": "M1 4AL",
    "outcode": "M1"
}
```

I think what this means for this PR is that I should close it. I don't think our solution is appropriate for the wider community and has fallen behind the currently documented API usage. For now we would prefer to stick with the single API request to get the Latitude and Longitude. If getAddress contacts us to advise us that this will change in the future then we will re-evaluate then.

My apologies for any effort you have expended on this.
All the best,
Mark

alexreisner:
That all makes sense, and I agree.
No worries at all. Thanks for the interest in contributing.

PR #1649, opened 2024-02-17
Update pg gem version in the Gemfile

takatea:
This PR updates the pg gem version in the Gemfile for developing this project.

## Details

While setting up my local environment for this project, I encountered an error during `bundle install` related to the installation of the pg gem:

```
pg_binary_decoder.c:115:20: error: call to undeclared function 'rb_tainted_str_new'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
```

This error was resolved by updating the pg gem to the latest version (1.5.x) from the old version (0.11.x). This allowed me to successfully run `bundle install`.
Additionally, I could run `bundle exec rake` and see all the tests pass.

How about this change?

## Environment

I have tested this change with the following environment:

### gcc

```
gcc -v
Apple clang version 14.0.3 (clang-1403.0.22.14.1)
Target: x86_64-apple-darwin22.5.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

### postgresql

```
psql --version psql
psql (PostgreSQL) 14.11 (Homebrew)
```

#### Ruby

```
ruby -v
ruby 3.3.0 (2023-12-25 revision 5124f9ac75) [x86_64-darwin22]
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,7 +24,7 @@ group :test do
   gem 'webmock'
 
   platforms :ruby do
-    gem 'pg', '~> 0.11'
+    gem 'pg', '~> 1.5'
     gem 'mysql2', '~> 0.5.4'
   end
 


alexreisner:
Thanks for this. And yes, I think it's safe to bump the pg version number. However, Geocoder currently supports Ruby >=2.1, and pg 1.4.5 requires Ruby 2.5+, so that's a problem. The latest version of pg that supports Ruby 2.1 is 1.1.4. Does pg v1.1.4 install correctly on your system? If so, let's bump it to that instead.

takatea:
@alexreisner 
Thank you for your comment!
I got the same error described the description when I tried to change the pg version to `1.1.4` and  then ran `bundle` install...

alexreisner:
Can you try troubleshooting your install of the pg gem, maybe by using the option discussed [here](https://stackoverflow.com/questions/77799851/client-c17824-error-call-to-undeclared-function-rb-tainted-str-new2-iso-c)? We can't bump the pg version past 1.1.4.

takatea:
@alexreisner 

Thanks!
I got that Ruby 2.1 need to be supported!
So, I am trying to change pg gem version to `=> 0.11` in Gemfile and trying to test with Ruby 2.1.

PR #1656, opened 2024-06-03
Update mapbox result to use place_name as address

tmh-dev:
Follow up to https://github.com/alexreisner/geocoder/pull/1655

Non-nil values are now being returned (when appropriate) for the `city`, `state`, `state_code`, `postal_code`, `country`, `country_code`, and `neighborhood` methods. However, as a result the `Geocoder::Result::Mapbox#address` method needs to be updated otherwise the returned value will contain duplicated text. 

Example: Given that the `city` method now returns a value for "place" type results, a search result for "Chicago, IL" will return "Chicago, Chicago, Illinois, United States". This is because `.place_name` and `.city` return the same value, and both are being used to build the address. 

To solve this, we can simply use the `place_name` property on the Mapbox payload as the address rather than building our own. 

Specs and fixtures have been added to test the different types of results you can get with Mapbox.

### DIFF
--- a/lib/geocoder/results/mapbox.rb
+++ b/lib/geocoder/results/mapbox.rb
@@ -56,7 +56,7 @@ module Geocoder::Result
     end
 
     def address
-      [place_name, street, city, state, postal_code, country].compact.join(', ')
+      data['place_name']
     end
 
     private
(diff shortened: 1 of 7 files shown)


xlozinguez (approved):

alexreisner:
Excellent, thanks!

PR #1667, opened 2024-11-11
Feature/azure maps

AhlOct:
- [x] Lookups
- [x] Results
- [x] Tests
- [x] Documentation


alexreisner:
Thanks for this. It looks great. Just one thing: there's no need to add the :limit option in configuration.rb. It should work fine without that. If you can remove that, I believe it will fix the erroring test. Thanks!

empeje (approved):

lib/geocoder/lookups/azure.rb
```diff
@@ -0,0 +1,56 @@
...
+    def query_url_params(query)
+      params = {
+        'api-version' => 1.0,
+        'language' => query.options[:language] || 'en',
+        'limit' => configuration.limit || 10,
+        'query' => query.sanitized_text,
+        'subscription-key' => configuration.api_key
+      }
+
+      params.merge!(super)
```
alexreisner, line 39: For readability, I would remove the `!` from the method name, since `params` is discarded when the method exits (the in-place merging is irrelevant, we're just returning the result of the merge).

lib/geocoder/lookup.rb
```diff
@@ -78,6 +79,7 @@ def ip_services
       @ip_services ||= [
         :baidu_ip,
         :abstract_api,
+        :azure,
```
alexreisner, line 82: Please remove this line. `:azure` is for physical addresses, not IP addresses.

alexreisner (changes requested):
Sorry, there were a few additional issues. In addition to what I've mentioned in comments:

1. You need to add a result fixture called `azure_madison_square_garden`. This is expected by the tests that run on all lookups to make sure global result attributes are supported.
2. After making all the recommended changes, you'll still see one failing test. The failure message is "Lookup azure does not return empty array when no results." Can you look into fixing that?

Thanks. And let me know if you have any questions about any comments.

AhlOct:
Thanks for your suggestions. I've fixed all issues you have mentioned and test 100% passed in my local. Now, I think we can run test again.

README_API_GUIDE.md
```diff
@@ -61,6 +61,19 @@ Global Street Address Lookups
       ```
     * Via environment variables and other external methods. See **Setting AWS Credentials** in the [AWS SDK for Ruby Developer Guide](https://docs.aws.amazon.com/sdk-for-ruby/v3/developer-guide/setup-config.html).
 
+### Azure (`:azure`)
+
+* **API key**: required (set `Geocoder.configure(lookup: :azure, api_key: "your_api_key", limit: your_limit)`)
```
alexreisner, line 66: The info in parentheses is duplicated below. Let's remove it here.

README_API_GUIDE.md
```diff
@@ -61,6 +61,19 @@ Global Street Address Lookups
...
+* **Key signup**: https://azure.microsoft.com/en-us/products/azure-maps
+* **Quota**: 5,000 request/month with free API key, more with paid keys (see https://azure.microsoft.com/en-us/pricing/details/azure-maps)
+* **Region**: world
+* **SSL support**: yes
+* **Languages**: see https://learn.microsoft.com/en-us/azure/azure-maps/supported-languages
+* **Documentation**: https://learn.microsoft.com/en-us/azure/azure-maps
+* **Terms of Service**: https://azure.microsoft.com/en-us/support/legal
+* **Limitations**: Azure Maps doesn't have any maximum daily limits on the number of requests that can be made, however there are limits to the maximum number of queries per second (QPS) (see https://learn.microsoft.com/en-us/azure/azure-maps/azure-maps-qps-rate-limits)
+* **Notes**: To use Azure, set `Geocoder.configure(lookup: :azure, api_key: "your_api_key", limit: your_limit)` :limit - restrict the maximum amount of returned results, e.g. limit: 10.
```
alexreisner, line 75: Two things:

1. Let's fix the syntax, showing that :limit has to be specified in an `azure: {...}` hash.
2. This makes it sound like :limit is required, which it isn't,, so let's specify the default (10) and mention that it's optional.

alexreisner (changes requested):
Great, thanks! Just one more thing, because I forgot to look at the `README`. Two minor changes there and then we're ready to merge.

README_API_GUIDE.md
AhlOct: I've updated README_API_GUIDE.md, sorry I forgot to update it yesterday.

alexreisner: No worries, thanks!

PR #1675, opened 2024-12-18
Fixed NoMethod error when querying with an instance of IPAddr

### LINKED ISSUE #1673: Does not work with IPAddr (since 1.8.4)
### Expected behavior

Geocoder should handle instances of `IPAddr` (used for database columns with type `inet`) - like mentioned in #1513.

After upgrading to 1.8.4 this does not work anymore. 
This regression was caused by fixing #1662 with 95eeaba15714ff57052abc89f2c54a3415fa6139.

### Actual behavior

The `sanitized_text` method tries to call `IPAddr#strip`, which does not exist:
https://github.com/alexreisner/geocoder/blob/v1.8.4/lib/geocoder/query.rb#L26

### Steps to reproduce

```
myapp(dev)> Geocoder::Query.new(IPAddr.new "1.1.1.1").sanitized_text
(myapp):8:in `<main>': undefined method `strip' for an instance of IPAddr (NoMethodError)

        text.strip
            ^^^^^^
```

This patch allows using `1.8.4` in my app:

```ruby
IPAddr.class_eval do 
  delegate :strip, to: :to_s
end
```

### Environment info

* Geocoder version: 1.8.4
* Rails version: 8
* Database (if applicable): 
* Lookup (if applicable): ipdata_co

mattlemx:
fixes #1673

### DIFF
--- a/lib/geocoder/query.rb
+++ b/lib/geocoder/query.rb
@@ -23,7 +23,7 @@ module Geocoder
           text.split(/\s*,\s*/).join(',')
         end
       else
-        text.strip
+        text.to_s.strip
       end
     end
 
--- a/test/unit/query_test.rb
+++ b/test/unit/query_test.rb
@@ -5,10 +5,14 @@ class QueryTest < GeocoderTestCase
 
   def test_detect_ipv4
     assert Geocoder::Query.new("232.65.123.94").ip_address?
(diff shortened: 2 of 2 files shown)


alexreisner:
Perfect, thanks!

PR #1679, opened 2025-01-30
Fix Undefined Variable Error in ip2location_io

14113:
This fixes a NameError caused by an undefined local variable or method (null) in the ip2location_io service:

```ruby
undefined local variable or method `null' for an instance of Geocoder::Lookup::Ip2locationIo (NameError)
```


alexreisner:
Thanks!

Issue #1657, opened 2024-06-12
Error when MapBox auth key is invalid


maxolasersquad:
### Expected behavior

No error should be thrown.

### Actual behavior

The following error is produced:

```
/usr/local/bundle/gems/geocoder-1.8.3/lib/geocoder/lookups/base.rb:46:in `search': undefined method `map' for false:FalseClass (NoMethodError)

        results(query).map{ |r|
                      ^^^^
Did you mean?  tap
```

### Steps to reproduce

Set up a Ruby on Rails application with Geocoder and configure it for MapBox, but with an invalid auth token. Try to save an Address model that uses Geocoder.

```
irb(main):001:0> a = Address.find 47
  Address Load (0.6ms)  SELECT "addresses".* FROM "addresses" WHERE "addresses"."id" = $1 LIMIT $2  [["id", 47], ["LIMIT", 1]]
=> 
#<Address:0x0000741c81b8f2b0
...
> a.save
  TRANSACTION (0.2ms)  BEGIN
Dalli::Server#connect 170.23.0.7:11211
Geocoding API error: 401 Unauthorized
  TRANSACTION (0.3ms)  ROLLBACK
/usr/local/bundle/gems/geocoder-1.8.3/lib/geocoder/lookups/base.rb:46:in `search': undefined method `map' for false:FalseClass (NoMethodError)

        results(query).map{ |r|
                      ^^^^
Did you mean?  tap
```

### Environment info

* Geocoder version: 1.8.3
* Rails version: 7.0.8
* Database (if applicable): Postgres
* Lookup (if applicable): MapBox

### More information

Geocoder configuration.

```ruby
Geocoder.configure(
  lookup: :mapbox,
  use_https: true,
  api_key: ENV['MAPBOX'],
  units: :mi,
)
```

ys7i:
@alexreisner I created a PR for this bug. https://github.com/alexreisner/geocoder/pull/1682

PR #1682, opened 2025-03-11
bug: Fixed unexpected error with invalid api key when exception-raising is not configured

ys7i:
This PR fix the open bug issue https://github.com/alexreisner/geocoder/issues/1657 .

## Bug
If geocoder is configured with MapBox and invalid api key and `always_raise` doesn't contain `Geocoder::InvalidApiKey`, `Geocoder.search` will throw the following error.
```
/usr/local/bundle/gems/geocoder-1.8.3/lib/geocoder/lookups/base.rb:46:in `search': undefined method `map' for false:FalseClass (NoMethodError)

        results(query).map{ |r|
                      ^^^^
Did you mean?  tap
```

## Why this happened
`results(query)` is evaluated as `false` in https://github.com/alexreisner/geocoder/blob/5b379a5f626bcd6b2e932627996ef6ba22761afc/lib/geocoder/lookups/base.rb#L46
because `raise_error` is called in the `results` method, and `results` returns `false` if `always_raise` doesn't contain the argument error.

### DIFF
--- a/lib/geocoder/lookups/mapbox.rb
+++ b/lib/geocoder/lookups/mapbox.rb
@@ -20,6 +20,7 @@ module Geocoder::Lookup
         sort_relevant_feature(data['features'])
       elsif data['message'] =~ /Invalid\sToken/
         raise_error(Geocoder::InvalidApiKey, data['message'])
+        []
       else
         []
       end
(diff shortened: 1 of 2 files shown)


alexreisner:
This is excellent, thanks!

PR #1686, opened 2025-06-01
Add IPinfo Lite service

iltempo:
IPinfo recently added a [Lite service](https://ipinfo.io/lite) that allows unlimited free IP lookups on the country level. This PR adds support for that service.

The results do not contain location information. This is why some tests are skipped.

### DIFF
--- a/README_API_GUIDE.md
+++ b/README_API_GUIDE.md
@@ -655,6 +655,19 @@ IP Address Lookups
 * **Documentation**: https://ipinfo.io/developers
 * **Terms of Service**: https://ipinfo.io/terms-of-service
 
+### IPInfo.io Lite (`:ipinfo_io_lite`)
+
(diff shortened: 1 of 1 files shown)


iltempo:
Thank you for the quick review and the feedback @alexreisner. I fixed the PR and added some general information about the service to a readme file, which I did not spot in the first place.

alexreisner:
And now it's been released! New version was long overdue. Sorry for the delay.

hemaldev:
@alexreisner Thank you, sir! No need to apologise. Thanks for your amazing work.

@iltempo Thanks for getting back to me and for providing this solution. After I submitted my comment I did think maybe a possible way forward would be to fork the repo and reference it in my Gemfile, similar to what you are suggesting. But @alexreisner beat us to it! (Out of curiosity, as you have forked this repo, do you plan on keeping your copy on your account or will you delete it now that your work has been merged? This would obviously have impacted us should we have referenced it).

PR #1687, opened 2025-06-05
update valid_region_codes for trimble to include new regions for Mexico and Middle East

x6iae:
According to the [trimble api documentation](https://developer.trimblemaps.com/restful-apis/location/single-search/single-search-api/#test-the-api-now) (which is also currently referenced in this codebase), There are now 2 additional regions supported which are currently not included in the set

The two additional ones are 
1. Mexico (MX)
2. Middle East (ME)

Making the full set to be the following:

1. North America (NA)
4. Europe (EU)
5. Asia (AS)
6. Africa (AF)
7. South America (SA)
8. Oceania (OC)
9. Mexico (MX)
10. Middle East (ME)

### DIFF
--- a/lib/geocoder/lookups/pc_miler.rb
+++ b/lib/geocoder/lookups/pc_miler.rb
@@ -10,10 +10,12 @@ module Geocoder::Lookup
       # AF: Africa
       # AS: Asia
       # EU: Europe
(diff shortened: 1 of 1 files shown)


alexreisner:
Thanks for this!

PR #1688, opened 2025-06-07
Return data['place_name'] for Geocoder:Result::Mapbox#place_name

### LINKED ISSUE #1661: Mapbox returning "text" instead of "place_name" key
### Expected behavior

When calling something "result.place_name", i expect the value of the "place_name" key.

### Actual behavior

The "place_name" method returns "data['text']" instead. Despite there being a "data['place_name']" available.

### Steps to reproduce

Do a Geocode for some long address. For example "Bahnhofsstraße 5 Passau".

https://www.rubydoc.info/gems/geocoder/Geocoder%2FResult%2FMapbox:place_name

### Environment info

* Geocoder version: 1.8.3
* Rails version: 7.1.3
* Lookup (if applicable): Mapbox

practicingdev:
Fixes #1661

<strike>I was not able to easily get the tests up and running but it looks like `test/fixtures/mapbox_chicago_il` has what is needed to verify the new behavior because it should return "Chicago, Illinois, United States" after this change, wheras before the change it will return "Chicago"</strike>

(I was able to get the single test file running but did not run the full suite, but it looks like I was able to get it to go green in all the relevant places)

### DIFF
--- a/lib/geocoder/results/mapbox.rb
+++ b/lib/geocoder/results/mapbox.rb
@@ -8,7 +8,7 @@ module Geocoder::Result
     end
 
     def place_name
-      data['text']
+      data['place_name']
     end
 
     def street
@@ -78,4 +78,3 @@ module Geocoder::Result
     end
   end
 end
-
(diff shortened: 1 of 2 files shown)


alexreisner:
This looks great, thank you!

Issue #1689, opened 2025-06-16
Unable to perform structured Nominatim lookup


janko:
I'm trying to perform a Nominatim lookup by passing the address via the `street`/`city`/`country`/`state`/`postalcode` query parameters, which is what Nominatim calls a structured lookup.

```rb
Geocoder.search(nil, params: {
  street: "...",
  city: "...",
  country: "...",
  state: "...",
  postalcode: "...",
})
```

Unfortunately, since Geocoder always passes the `q` query parameter, regardless of whether it's blank, I get an error from Nominatim:

```
Structured query parameters(amenity, street, city, county, state, postalcode, country) cannot be used together with 'q' parameter.
```

I was considering opening a PR with this change:

```diff
diff --git a/lib/geocoder/lookups/nominatim.rb b/lib/geocoder/lookups/nominatim.rb
index f7e119a..b15c089 100644
--- a/lib/geocoder/lookups/nominatim.rb
+++ b/lib/geocoder/lookups/nominatim.rb
@@ -55,7 +55,7 @@ module Geocoder::Lookup
         lat,lon = query.coordinates
         params[:lat] = lat
         params[:lon] = lon
-      else
+      elsif !query.params_given?
         params[:q] = query.sanitized_text
       end
       params
diff --git a/lib/geocoder/query.rb b/lib/geocoder/query.rb
index c509a7b..3dc15c2 100644
--- a/lib/geocoder/query.rb
+++ b/lib/geocoder/query.rb
@@ -116,8 +116,6 @@ module Geocoder
       options[:language]
     end
 
-    private # ----------------------------------------------------------------
-
     def params_given?
       !!(options[:params].is_a?(Hash) and options[:params].keys.size > 0)
     end
```

But I realized that maybe not all `:params` apply to structured lookup, some could be valid together with the `q` parameter.

Any idea how to solve / work around this?

janko:
I think additionally checking that query text is blank should be sufficiently safe, so I opened a PR.

janko:
It would probably be ergonomical to make the first argument to `Geocoder.search` optional, but didn't want to jump the horse 😃

PR #1690, opened 2025-06-16
Allow structured Nominatim lookup

janko:
The following currently fails with an error:

```rb
Geocoder.search(nil, params: {
  street: "...",
  city: "...",
  country: "...",
  state: "...",
  postalcode: "...",
})
```
```
Structured query parameters(amenity, street, city, county, state, postalcode, country) cannot be used together with 'q' parameter.
```

We fix this by skipping including the `q` parameter in case structured query parameters are passed and query text would be blank.

Closes https://github.com/alexreisner/geocoder/issues/1689

### DIFF
--- a/lib/geocoder/lookups/nominatim.rb
+++ b/lib/geocoder/lookups/nominatim.rb
@@ -55,7 +55,7 @@ module Geocoder::Lookup
         lat,lon = query.coordinates
         params[:lat] = lat
         params[:lon] = lon
-      else
+      elsif !query.params_given? || !query.sanitized_text.empty?
         params[:q] = query.sanitized_text
       end
       params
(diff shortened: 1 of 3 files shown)


alexreisner:
@janko thanks for this. I think your approach is good, just a few thoughts:

1. The purpose of your test is not clear to me. Perhaps it should be named something like `test_q_is_empty_with_structured_lookup`? It doesn't matter so much that you get a result back (since it's a fake result). Maybe the assertion should involve a `Query` object?
2. Are you sure the condition you've added (`!query.params_given? || !query.sanitized_text.empty?`) isn't inverted? It seems to me that q should be blank if there are no params given (but I might be confused).

PR #1691, opened 2025-06-23
Radius column can be an instance of Arel.sql

TheRusskiy:
With older Rails versions and Geocoder you could do something like:
```ruby
Person.near(city.coordinates, :"COALESCE(distance_preference, 100)")
# or
Person.near(city.coordinates, :"CASE WHEN travels_by_foot THEN foot_travel_distance ELSE car_travel_distance END")
```

But with newer Rails versions (well, by newer I mean like 6.1 or something) the query ends up looking like

```sql
SELECT MAX("COALESCE(distance_preference, 100)") FROM people
```

(there's no column `"COALESCE(distance_preference, 100)"`)

So usually within Rails you would do `Person.maximum(Arel("COALESCE(distance_preference, 100)"))`

But with Geocoder due to the `radius.is_a?(Symbol)` check it's not possible,
so with this PR I am extending the logic to work with Arel SQL literals.

After this change the following code works:
```ruby
Person.near(city.coordinates, Arel.sql("COALESCE(distance_preference, 100)"))
# or
Person.near(city.coordinates, Arel.sql("CASE WHEN travels_by_foot THEN foot_travel_distance ELSE car_travel_distance END"))
```

### DIFF
--- a/lib/geocoder/stores/active_record.rb
+++ b/lib/geocoder/stores/active_record.rb
@@ -132,7 +132,8 @@ module Geocoder::Store
         # If radius is a DB column name, bounding box should include
         # all rows within the maximum radius appearing in that column.
         # Note: performance is dependent on variability of radii.
-        bb_radius = radius.is_a?(Symbol) ? maximum(radius) : radius
+        radius_is_column = radius.is_a?(Symbol) || (defined?(Arel::Nodes::SqlLiteral) && radius.is_a?(Arel::Nodes::SqlLiteral))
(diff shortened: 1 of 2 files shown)


alexreisner:
This looks great, thanks! Can you just take a look at [the failing test](https://github.com/alexreisner/geocoder/actions/runs/15831024003/job/44798893133?pr=1691)?

alexreisner:
Yep, that fixed it. Thanks!

PR #1694, opened 2025-08-28
Update domain for ban_data_gouv_fr adapter

Intrepidd:
The official domain has changed a while ago, the old one was still working but yesterday there was an incident affecting only the old domain. It's back up again but it is probably best to use the official domain.

Reference: https://geoservices.ign.fr/documentation/services/services-geoplateforme/geocodage

### DIFF
--- a/lib/geocoder/lookups/ban_data_gouv_fr.rb
+++ b/lib/geocoder/lookups/ban_data_gouv_fr.rb
@@ -18,7 +18,7 @@ module Geocoder::Lookup
 
     def base_query_url(query)
       method = query.reverse_geocode? ? "reverse" : "search"
-      "#{protocol}://api-adresse.data.gouv.fr/#{method}/?"
+      "#{protocol}://data.geopf.fr/geocodage/#{method}/?"
     end
 
     def any_result?(doc)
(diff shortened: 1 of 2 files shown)


alexreisner:
Excellent, thanks!

sallesma:
@Intrepidd I started to have a lot errors after upgrading the geocoder gem.
My calls to the BAN triggered this error:
```
> Geocoder.search("_valid address_", lookup: :ban_data_gouv_fr)
(cyke):1:in '<main>': Geocoder::LookupTimeout (Geocoder::LookupTimeout)
.rbenv/versions/3.4.1/lib/ruby/gems/3.4.0/gems/net-http-0.6.0/lib/net/http.rb:1665:in 'TCPSocket#initialize': Failed to open TCP connection to data.geopf.fr:80 (execution expired) (Net::OpenTimeout)
```

I tried again with `use_https: true` in my config and it works well.
Maybe we should force https when using the BAN ?

sallesma:
I've released a new version with `use_https: true` a few minutes ago and I'm monitoring if this was the real cause of the issue. If it is, I'll open a PR to force https when using the BAN

PR #1697, opened 2025-10-07
Force https on BAN lookup

sallesma:
Following https://github.com/alexreisner/geocoder/pull/1694 (more info written there)

The BAN endpoint does not respond to http requests on port 80, which triggers a timeout. It works well when using https.

So I think we should force https when using the BAN instead of relying on the config


alexreisner:
I agree, thank you!

PR #1705, opened 2026-03-26
Update build

mikehale:
* Add ruby 4 to the build matrix
* Update the checkout action to the latest version
* Sort gemfile
* Add rdoc and benchmark (required for ruby >= 4.0)


alexreisner:
Thanks for this! Could you take a look at the test failure?

mikehale:
Apparently the latest JSON parser is much more strict about invalid escape codes in strings. The tests were failing under ruby 4 because the parser was not able to handle `\xC2\xA9`, so I updated the fixture with `\u00a9`.

alexreisner:
Good work! Thanks!

mikehale:
For posterity this is the changelog item when the json gem became more strict https://github.com/ruby/json/blob/master/CHANGES.md#2025-12-03-2170

Issue #1299, opened 2018-06-08
Exception thrown when using mapbox and query contains semicolons


12eh:
Mapbox interprets this as multiple queries and returns an array of results. `Geocoder::Lookup::Mapbox` expects it to be a hash, and when it tries a key lookup this throws `TypeError: no implicit conversion of String into Integer`.

### Expected behavior
This is definitely debatable, which is why I'm opening an issue rather than a PR right away. The possibilities I can think of are:
- throw an `ArgumentError` or something with an error message saying multiple-location queries aren't supported
- strip the semicolons out of the query before sending to mapbox (on the off chance that someone's query just happens to contain a semicolon and they don't intend for it to mean multiple separate queries)
- return the result for the first of the queries
- return the highest relevance result out of all the queries
- add support in geocoder for this concept and return an array of results as mapbox does

### Actual behavior
Throwing an exception which doesn't describe what the real issue is

### Steps to reproduce
```
> Geocoder.search('New York, NY;Los Angeles, CA')
TypeError: no implicit conversion of String into Integer
from /app/vendor/bundle/ruby/2.5.0/gems/geocoder-1.4.5/lib/geocoder/lookups/mapbox.rb:20:in `[]'
```

### Environment info

* Geocoder version: 1.4.5
* Rails version: 5.1.4
* Database (if applicable): 
* Lookup (if applicable): mapbox

alexreisner:
Very interesting--thanks for raising this issue! There has been an aborted attempt to add batch geocoding (see discussion at #1046)--maybe this is the time to try again. However, that's a much larger issue and should be handled separately. For now, here's what I think: if Mapbox provides a way to escape a semicolon (so it's interpreted literally, within the address, rather than as a record separator) then I think we should escape them. If not, then maybe truncating the query at the first semicolon makes the most sense (the logic being if Geocoder doesn't support batch queries then we shouldn't make Mapbox run one).

Any support we add for batch queries would require users to put addresses in an array, not a semicolon-delimited string, so this solution is relevant whether or not we implement batch geocoding.

Does that all make sense? Open to discussion.

12eh:
I don't see anything on https://www.mapbox.com/api-documentation/#geocoding mentioning a way to escape a semicolon (or mentioning semicolons in general). I think truncating the query at the first semicolon is a reasonable solution.

For our application, we're currently just removing all semicolons from the query before even passing to Geocoder, so I have no strong preference for how it's handled at this level.

PR #1709, opened 2026-06-05
Add Mapbox Geocoding v6 lookup

### LINKED ISSUE #1708: Mapbox: support the new Geocoding v6 API
## Summary

The current Mapbox lookup (`lib/geocoder/lookups/mapbox.rb`) uses the Geocoding v5 API:

```
https://api.mapbox.com/geocoding/v5/{dataset}/{query}.json
```

Mapbox has released the [Geocoding v6 API](https://docs.mapbox.com/api/search/geocoding/), and v5 is now documented as the "previous version". While v5 still works for address/place searches, it is effectively in legacy status:

- POI data has been removed from v5 (users are directed to the Search Box API).
- New features are only available in v6: unit-level geocoding, structured input, Smart Address Match, improved batch geocoding (same per-request price as v5).

## API differences

v6 changes both the endpoints and the response format, so it cannot be supported by just swapping the URL:

| | v5 | v6 |
|---|---|---|
| Forward | `GET /geocoding/v5/{dataset}/{query}.json` | `GET /search/geocode/v6/forward?q={query}` |
| Reverse | `GET /geocoding/v5/{dataset}/{lon},{lat}.json` | `GET /search/geocode/v6/reverse?longitude={lon}&latitude={lat}` |
| Permanent storage | separate dataset (`mapbox.places-permanent`) | `permanent=true` parameter |
| Response | feature with `place_name`, `text`, flat `context` array keyed by `id` prefix | feature with `properties.full_address`, `properties.name`, structured `properties.context` object (`country`, `region`, `place`, `postcode`, `street`, `address`, ...) |

Because the response structure is different, `Geocoder::Result::Mapbox` (which parses `place_name` / `text` / `context[].id`) does not work with v6 responses.

## Proposal

Add a new `:mapbox_v6` lookup (with its own result class) alongside the existing `:mapbox` lookup, following the same pattern as other versioned lookups. This keeps existing users on v5 unaffected while allowing opt-in to v6.

I'm happy to submit a PR for this if the approach sounds reasonable.

## References

- Geocoding v6 docs: https://docs.mapbox.com/api/search/geocoding/
- Geocoding v5 docs (previous version): https://docs.mapbox.com/api/search/geocoding-v5/

takatea:
Closes #1708

## Summary

Adds opt-in support for the [Mapbox Geocoding v6 API](https://docs.mapbox.com/api/search/geocoding/) to the existing `:mapbox` lookup:

```ruby
Geocoder.configure(mapbox: {version: 6})
```

v5 remains the default, so existing users are unaffected. Since the public result interface is identical between the two versions, this switches versions via configuration within the one lookup rather than registering a separate `:mapbox_v6` lookup.

## Design references

The design follows existing patterns in this gem:

- **Version switch via a lookup option**: modeled on the Mapquest lookup's `:open` option, which switches between the licensed and open endpoint families within one lookup ([`lib/geocoder/lookups/mapquest.rb#L19`](https://github.com/alexreisner/geocoder/blob/aa134e37153fe2ac44b071ae2eaeb266383efe66/lib/geocoder/lookups/mapquest.rb#L19), introduced in #315).
- **Separate result class per response format**: the v6 response is parsed by its own `Geocoder::Result::MapboxV6` class, both because the response shapes are genuinely different and to keep the v5 code easy to delete later. When Mapbox eventually sunsets v5, removing it amounts to dropping the `else` branches and `results/mapbox.rb` — similar to how the HERE lookup was migrated to API v7 in place (#1575, released in 1.8.0 with a CHANGELOG note).
- **`#precision` accessor**: the v6 result exposes `properties.coordinates.accuracy` (`rooftop` / `parcel` / `interpolated` / ...) as `#precision`, following `Geocoder::Result::Google#precision` which returns the semantically equivalent `location_type`.

## API differences handled

| | v5 (default) | v6 (`version: 6`) |
|---|---|---|
| Forward | `GET /geocoding/v5/{dataset}/{query}.json` | `GET /search/geocode/v6/forward?q={query}` |
| Reverse | same endpoint, `{lon},{lat}.json` in path | `GET /search/geocode/v6/reverse?longitude=&latitude=` |
| Permanent storage | separate dataset (`mapbox.places-permanent`) | `permanent=true` request param |
| Address string | `place_name` | `properties.full_address` |
| Context | flat array matched by `id` prefix | structured `properties.context` object |
| State code | `properties.short_code` (`US-NY`, needs splitting) | `context.region.region_code` (`NY`) |
| Country code | `properties.short_code` | `context.country.country_code` |

## Changes

- `lib/geocoder/lookups/mapbox.rb` — `version: 6` branches for endpoint and params; v5 paths untouched. Same `InvalidApiKey` handling and semicolon truncation (#1299) apply to both versions
- `lib/geocoder/results/mapbox_v6.rb` — parses the structured `properties.context` object; adds a `#precision` accessor for the v6 accuracy value
- tests + fixtures following the existing `mapbox_*` patterns (`test_helper.rb` picks the fixture prefix by configured version)
- `README_API_GUIDE.md` — documents the `version` option under the existing Mapbox section

## Verified with live API

Ran the same queries through v5 and v6 with a real access token to confirm the result interfaces line up.

```
### query: "1600 Pennsylvania Avenue NW, Washington, DC"
v5 (Geocoder::Result::Mapbox):
  coordinates  [38.897684, -77.036574]
  address      "1600 Pennsylvania Avenue Northwest, Washington, District of Columbia 20500, United States"
  street       nil
  city         "Washington"
  state        "District of Columbia"
  state_code   "DC"
  postal_code  "20500"
  country      "United States"
  country_code "US"
  neighborhood "National Mall"
v6 (Geocoder::Result::MapboxV6):
  coordinates  [38.897684, -77.036574]
  address      "1600 Pennsylvania Avenue Northwest, Washington, District of Columbia 20500, United States"
  street       "1600 Pennsylvania Avenue Northwest"
  city         "Washington"
  state        "District of Columbia"
  state_code   "DC"
  postal_code  "20500"
  country      "United States"
  country_code "US"
  neighborhood "National Mall"
  precision    "rooftop"

### query: [40.750755, -73.993710125]  (reverse)
v5 (Geocoder::Result::Mapbox):
  coordinates  [40.750711, -73.993849]
  address      "445 Fashion Ave, New York, New York 10121, United States"
  street       nil
  city         "New York"
  state        "New York"
  state_code   "NY"
  postal_code  "10121"
  country      "United States"
  country_code "US"
  neighborhood "Chelsea"
v6 (Geocoder::Result::MapboxV6):
  coordinates  [40.750711, -73.993849]
  address      "445 Fashion Ave, New York, New York 10121, United States"
  street       "445 Fashion Ave"
  city         "New York"
  state        "New York"
  state_code   "NY"
  postal_code  "10121"
  country      "United States"
  country_code "US"
  neighborhood "Chelsea"
  precision    "rooftop"
```

Notes from live verification:

- The two versions return identical values for every shared field; v6 additionally fills `street` (v5's `street` reads `properties.address`, which the live v5 API no longer populates for these queries) and exposes `precision`.
- A POI query ("Madison Square Garden, New York, NY") resolves to a neighborhood on **both** versions — POI data has been removed from v5 as well, so this is not a v6 regression.
- An invalid token returns `Not Authorized - Invalid Token` on v6 too, so the existing `Geocoder::InvalidApiKey` handling works unchanged.

## Tests

```
$ bundle exec rake  # 613 tests, 3643 assertions, 0 failures, 0 errors
```


alexreisner:
Thanks for this great PR! The code looks good with one exception. APIs change all the time, and one goal of Geocoder is to save users from having to think about those changes, while allowing them to use the latest/recommended API versions. So I think the default behavior for people using the `:mapbox` lookup should be for them to be silently upgraded to v6, assuming there's nothing that could break their apps. If there is any reason people might want to stick with v5, then we can have a separate v5 lookup (eg, `mapbox_v5`), but generally I only do that when there are compelling reasons to stick with an old version of the API long term. I haven't dug too far into the v6 changes--is there any reason to continue supporting v5 long-term, in your mind?

takatea:
@alexreisner Thank you, this makes a lot of sense to me! I agree with the philosophy that `:mapbox` users should be upgraded to v6 transparently.

To answer your question directly: I do not see a compelling reason to keep v5 supported long-term. In the live comparison, v6 returned identical values for every shared field, and it additionally fills `street` and exposes `precision`, so the result interface is effectively a superset of v5. I could not find any data that v5 returns but v6 does not. I am therefore happy to make v6 the default and remove v5 in place (following the HERE v7 approach from #1575), along with a CHANGELOG note.

There is one migration detail that I would like your call on, since it is the only thing I can see that might silently break existing apps:

**Permanent storage.** In v5, this is configured through the dataset:

```ruby
Geocoder.configure(mapbox: { dataset: "mapbox.places-permanent" })
```

In v6, the dataset concept is gone and a request param is used instead, which already works today through Geocoder's generic `params` option:

```ruby
Geocoder.search(address, params: { permanent: true })
```

The concern is that anyone currently relying on `dataset: "mapbox.places-permanent"` would, after a silent upgrade, start making non-permanent requests without noticing, which has billing/ToS implications for them.

How do you think we should handle this? Personally, I lean toward simply documenting it with a CHANGELOG / README migration note. That said, if you would prefer to emit a temporary deprecation instead, I am happy to map a permanent `dataset` setting to `permanent: true` for one release with a warning, and then drop it later. I can implement it either way, so I would appreciate your opinion before I update the PR.

alexreisner:
@takatea great question (and sorry for the delay). Let's handle it like this:

- add a note in the CHANGELOG (I can do this when it's created)
- if someone has `{ dataset: "mapbox.places-permanent" }`, automatically merge `{ permanent: true }` into `params` and print a deprecation warning

That way nobody's app behavior will change, and people will hopefully adjust their code. We could probably leave it like this for several versions. Does that make sense to you?

Issue #1670, opened 2024-11-20
Geocoder::CacheStore::Generic.new(Rails.cache, {})) does not support expiration


Saoma1:
Hello 👋 

I've encountered an issue when using Geocoder::CacheStore::Generic.new(Rails.cache, {}) and passing an expiration in the cache_options. The @cache_hit is always true (after the first request).

Upon reviewing the implementation [here](https://github.com/alexreisner/geocoder/blob/master/lib/geocoder/cache_stores/generic.rb), it appears that expires_in is never passed to Rails.cache, which handles expiration logic.

I came across a related [issue](https://github.com/alexreisner/geocoder/pull/1560), but it focuses mainly on updating the documentation. Would it be possible to implement a proper Geocoder::CacheStore::RailsCache? This would eliminate the need for users to extend or monkey patch the gem.

For context, here is my configuration:

```ruby
Geocoder.configure(
  timeout: 3, # geocoding service timeout (secs)
  lookup: :google, # name of geocoding service (symbol)
  language: :de, # ISO-639 language code
  use_https: true, # use HTTPS for lookup requests? (if supported)
  api_key: "key", # API key for geocoding service
  cache: Geocoder::CacheStore::Generic.new(Rails.cache, {}),
  cache_options: {
    expiration: 2.seconds
  },
  units: :km, # :km for kilometers or :mi for miles
)
```

Thank you for considering this! 😊

alexreisner:
Good point. I think this was done because we didn't want to make it seem like expiration is supported by cache back-ends that don't actually support it. But yes, I'm open to a PR For a proper Rails cache solution. (Thanks!)

DeedleFake:
I just ran into this. The Rails cache in the project I'm working on at work is backed by Redis, but the redirect through the Rails cache causes the `expiration` config to get silently ignored, and I only found out because there is now gigabytes of data cached in Redis. Worse, the documentation does not at all explain how to configure the cache well. Reading the actual code, it seems that the cache backend is chosen _based on the class name of the value given for the `cache` in the config_, which means that if it's not a class exactly named `Redis` that is not in a module in a module then you get the generic, no expiration cache layer. At the very least the documentation needs to be improved.

alexreisner:
This should be fixed by #1711 but please re-open if you're still having this issue.

PR #1711, opened 2026-08-07
Allow configuring the cache with a cache store instance

pioz:
## Problem

`Geocoder::Cache` picks its cache store adapter from the **class name** of the configured `cache:` object:

```ruby
@class = (Geocoder::CacheStore.const_get("#{store.class}", false) rescue Geocoder::CacheStore::Generic)
```

Any object whose class is not literally named `Redis` silently falls back to `Geocoder::CacheStore::Generic`, which does not apply `cache_options[:expiration]`. In our app the Redis client is wrapped in a small namespacing class, so the configured `expiration: 2.days` was silently ignored and keys accumulated in Redis with no TTL — we only noticed when Redis memory kept growing. The same pitfall is described in #1670 ([comment](https://github.com/alexreisner/geocoder/issues/1670#issuecomment-2699295022)).

Passing an already-built cache store instance — as the README suggests for Rails (`cache: Geocoder::CacheStore::Generic.new(Rails.cache, {})`) — only works by accident today: `Geocoder::Cache` wraps the instance in a *second* `Generic` adapter, whose duck-typing dispatch happens to delegate `write`/`read` to the inner adapter. `expire` is actually broken in that setup, because `Generic#remove` calls `store.delete`, which cache store adapters don't implement:

```
NoMethodError: undefined method 'delete' for an instance of Geocoder::CacheStore::Generic
```

## Change

When the configured `cache:` is already a `Geocoder::CacheStore::Base` instance, use it directly instead of wrapping it. This gives an explicit way to choose the adapter for stores whose class name is not recognized:

```ruby
Geocoder.configure(
  cache: Geocoder::CacheStore::Redis.new(MyRedisWrapper.new, {prefix: "geocoder:", expiration: 2.days})
)
```

Expiration support is still never guessed for unrecognized stores (in line with the intent expressed in #1670): the user opts in explicitly by picking the adapter.

Backward compatible: the class name lookup and the `Generic` fallback are unchanged for every other value of `cache:`. The README example for `Rails.cache` now also works fully (no double wrapping, `expire` included).

Related to #1670.


alexreisner:
This is excellent. Thank you for the great code and clear explanation. Merging!

pioz:
Yeah! Any change to get a new geocoder release (1.8.7) with this update?

